PR #1, opened 2018-01-17
[MCOMPILER-320] Introduce additionalCompilePathItems property

dmlloyd:


dmlloyd:
Any updates on this? @rfscholte ?

gastaldi:
@rfscholte any chance to get this merged anytime soon? This is really needed in order to avoid using the forked JBoss maven-compiler-plugin. Thanks!

eolivelli:
@dmlloyd  @gastaldi @rfscholte 
I have rebased onto master and push forced to
https://github.com/apache/maven-compiler-plugin/tree/MCOMPILER-320

Let's wait for CI
https://builds.apache.org/job/maven-box/job/maven-compiler-plugin/job/MCOMPILER-320/48/

Did you have comments about this change @rfscholte  ?
I see this branch that seems to relate to MCOMPILER-320
https://github.com/apache/maven-compiler-plugin/commits/MCOMPILER-320_sunmisc

eolivelli:
@rfscholte it looks like the maven compiler plugin is not in good shape.
We should fix it before committing this patch

eolivelli:
Thank you @rfscholte for giving me the pointer, I should have read the full JIRA before jumping on this train.
In the mid term we want to be able to let Maven understand dependencies between mojos/tasks execution (input/output). 
I agree that we should leverage as most as possible the Project Model and so the 'dependecies' feature.
With this change we will add out-of-band dependencies additions and it smells like an hack.

Initially I was driven in my LGTM by the needs of @dmlloyd but we should think about the future and every feature we add now we have to maintain it for ever.
I am sorry that surefire already provides this feature.

If I understand correctly we should try to convert the original problem to an integration test and see if there is really no other way to achieve the same goal.

I don't know if is possible to build an 'extension' that changes the dependencies at build time instead of forking the plugin

ge0ffrey:
Any change on this? The lack of this PR getting merged is causing issues for OptaPlanner.

rfscholte:
@ge0ffrey I don't want to fix this by adding the `additionalClasspathElements` property. Any suggestion based on`<dependency/>` based is appreciated.

jira-importer:
Resolve #536

PR #3, opened 2018-02-27
[MCOMPILER-205] fixes incremental compilation

monperrus:
Fixes https://issues.apache.org/jira/browse/MCOMPILER-205.

* When useIncrementalCompilation=false, always recompile all source files.
* When useIncrementalCompilation=true (default), only recompiles stale sources.


src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -750,36 +750,16 @@ else if ( CompilerConfiguration.CompilerReuseStrategy.ReuseSame.getStrategy().eq
 
         IncrementalBuildHelperRequest incrementalBuildHelperRequest = null;
 
-        if ( useIncrementalCompilation )
+        if ( !useIncrementalCompilation )
         {
             getLog().debug( "useIncrementalCompilation enabled" );
```
aikar: You forgot to invert these log lines.

monperrus: good point! fixed.

yeroc:
What needs to be done to move this forward?

eolivelli:
Hi
did you run the integration tests suite?

mvn verify -Prun-its


Btw actually in 3.8.0 you just have to turn incremental compilation off in order to build only stale sources.
This behaviour is very error prone and it is to be used with caution.

With you fix it seems that you are dropping the part that is detecting stale files and inverting the meaning of the flag.
Can you explain more?

monperrus:
the fix does not drop anything, it indeed inverts the flag so that the behavior or the flag does what the meaning suggests, https://issues.apache.org/jira/browse/MCOMPILER-205

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -775,36 +775,16 @@ else if ( CompilerConfiguration.CompilerReuseStrategy.ReuseSame.getStrategy().eq
...
 
                 sources = getCompileSources( compiler, compilerConfiguration );
-                
+                compilerConfiguration.setSourceFiles( sources );
                 preparePaths( sources );
-
-                incrementalBuildHelperRequest = new IncrementalBuildHelperRequest().inputFiles( sources );
-
-                // CHECKSTYLE_OFF: LineLength
-                if ( ( compiler.getCompilerOutputStyle().equals( CompilerOutputStyle.ONE_OUTPUT_FILE_FOR_ALL_INPUT_FILES ) && !canUpdateTarget )
```
eolivelli: It seems to me that here we are losing this test.

monperrus: this simplification is on purpose for regular behavior and maintainability: we are not using incremental compilation, so we consider all sources.

the else branch is responsible for detecting stale changes.

rfscholte:
> What needs to be done to move this forward?

Did you read the comments of https://issues.apache.org/jira/browse/MCOMPILER-205 ? So in general it works as designed.
The root cause is a missing feature of https://codehaus-plexus.github.io/plexus-compiler/ where you can define is a specific compiler implementation has incremental support. Once that is available we can start applying it to this plugin.

yeroc:
@rfscholte Thanks for the update.  There's quite a bit of conflicting information on the ticket so for an outsider it's difficult to discern what's correct and what isn't.  That said, I scanned through the ticket again and couldn't find a reference to a missing feature in the compiler.  Can you elaborate on that or post a link to the relevant ticket for that?

rfscholte:
@yeroc nobody created an issue for it yet, but you might conclude it based on some comments. Both this PR and MCOMPILER-205 have a huge amount of comments, which makes it kind of hard to see the whole picture. 
Let's start by closing this PR, I won't accept it since reversing the logic is not the solution.

https://stackoverflow.com/questions/16963012/maven-compiler-recompile-all-files-instead-modified/49700942#49700942 is the best summary (and the accepted answer in incorrect)

cowwoc:
Who will create the new issue? Be sure to link the new issue to this one so we can follow. Thanks!

PR #12, opened 2019-02-20
Improve "Older projects with module-info" chapter

oehme:
The previous suggestion lead to all source code being compiled twice, once with Java 9 and once with Java 6. This is unnecessarily slow. The new suggestion first compiles everything with Java 6 (except the module-info file) and then compiles *only* the module-info file with Java 9. Since all the other files are already compiled at that point, javac will be able to reference them from the classpath. This significantly speeds up compilation.

I've also put the "minimum Java 6" section at the top, as that is the more common case. Instructions for Java 5 are now at the bottom of the page. This makes it more likely for people to copy-paste only what they really need.

       pulling in other changes.
 - [ ] Format the pull request title like `[MPH-XXX] - Fixes bug in ApproximateQuantiles`,
       commit message.

### DIFF
--- a/src/site/apt/examples/module-info.apt.vm
+++ b/src/site/apt/examples/module-info.apt.vm
@@ -33,64 +33,6 @@ Older projects with module-info
   <<<module-info.java>>> must be compiled with <<<release=9>>>, while the rest of the sources must be compiled with a
   lower version of <<<source>>>/<<<target>>>.
 
-  The preferred way to do this is by having 2 execution blocks are described below. JDK 9 only supports compilations for 
-  Java 6 and above, so projects wanting to be compatible with Java 5 or below need to use two different JDKs. With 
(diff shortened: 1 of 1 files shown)


sormuras:
Thanks for the PR, @oehme 

Do you have an integration test (or demo project) that uses this "split" compile strategy? I'm asking because when I did something similar, `javac` wasn't able to compile a sole `module-info.java` file that referred to already compiled packages and types.

rfscholte:
AFAIK the current setup is the only way to ensure that content of the classfiles and its modueldescriptor are reliable, i.e. all required modules are defined, all exported packages exist, all services and providers exist. An integration test should confirm this.

oehme:
I've tried this out on the slf4j project, because they had this exact setup. I flipped it around as suggested here and it compiled just fine. But:

> AFAIK the current setup is the only way to ensure that content of the classfiles and its modueldescriptor are reliable, i.e. all required modules are defined,

This is a problem indeed. Closing the PR.

PR #13, opened 2019-03-12
Emit a warning instead of throwing an exception

sormuras:
Prior this commit, an UnsupportedOperationException was thrown when a test module descriptor is present and a main module descriptor is absent.

Now only a warning is emitted informing the user to either fix the modular setup or to allow users ignoring the warning. They might want to test that specific setup after all.

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/TestCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/TestCompilerMojo.java
@@ -369,9 +369,11 @@ public class TestCompilerMojo
                 // very odd
                 // Means that main sources must be compiled with -modulesource and -Xmodule:<moduleName>
                 // However, this has a huge impact since you can't simply use it as a classpathEntry 
-                // due to extra folder in between
-                throw new UnsupportedOperationException( "Can't compile test sources "
-                    + "when main sources are missing a module descriptor" );
+                // due to extra folder in between ... or the user wants to test his main code on the
+                // classpath used by a named module ... or ...
+                getLog().warn( "Modular black-box test compilation with no main module?!" );
+
+                // TODO classpathElements = List.of( mainOutputDirectory );
             }
         }
         else


src/main/java/org/apache/maven/plugin/compiler/TestCompilerMojo.java
```diff
@@ -369,9 +369,11 @@ else if ( Double.valueOf( getTarget() ) < Double.valueOf( MODULE_INFO_TARGET ) )
...
                 // Means that main sources must be compiled with -modulesource and -Xmodule:<moduleName>
                 // However, this has a huge impact since you can't simply use it as a classpathEntry 
-                // due to extra folder in between
-                throw new UnsupportedOperationException( "Can't compile test sources "
-                    + "when main sources are missing a module descriptor" );
+                // due to extra folder in between ... or the user wants to test his main code on the
+                // classpath used by a named module ... or ...
+                getLog().warn( "Modular black-box test compilation with no main module?!" );
+
+                // TODO classpathElements = List.of( mainOutputDirectory );
```
eolivelli: Todo? Can you link a JIRA? In order notbto forget?

sormuras: > "In order not to forget"

This. I'm not sure what to do here, actually. That's why I implemented the `throws` variant first. Now, a warning is emitted, and any subsequent "mis-configuration" will lead to an error. Presumably when calling `javac`... but, as stated in the comment, this "main=classic + test=module" case is rare and not well understood.

sormuras:
> So you want to compile your codebase in classpath mode and run tests as a module which is consuming an unnamed module (classpath based)?

Exactly. Similar to this setup https://github.com/micromata/sawdust/tree/master/modular-blackbox-mono -- but with a "classic" `src/main/java` directory.

src/main/java/org/apache/maven/plugin/compiler/TestCompilerMojo.java
rfscholte: @sormuras you know what I'm going to say ;) documentation is key here. Without having clear examples what's happening here and why somebody wants to do this I'd prefer to keep the Exception.

sormuras: I already excluded that case from an example project: https://github.com/micromata/sawdust/blob/master/pom.xml#L15

But, I might have found a valid use-case today: say you want to test an external module (on the module path) without writing main code on your own.

sormuras:
Or just have a standalone modular test checking "external" modules from the JDK and other test dependencies.

sormuras:
Those lines should allow that use-case already.
```
                // No main binaries available? Means we're a test-only project.
                if ( !mainOutputDirectory.exists() )
                {
                    return;
                }
```
 Strange...

sormuras:
...`mvn clean` solved the errors on my side.

Having the exception thrown if `target/classes` is present **without** containing a `module-info.class` and also having `target/test-classes` **with** a `module-descriptor` present at the same time ... is suspicious.

Tibor17:
@sormuras your last comment makes sense to me but @rfscholte is an expert so he has to say if we are right.

PR #19, opened 2019-05-21
(doc) Simplify CompilationFailureException

rhowe:
This class can be a little bit more concise.

The nullity check in `longMessage()` is made redundant by the lack of an equivalent check in `shortMessage`. `CompilationFailureException` is thrown in two places in `AbstractCompilerMojo` and both of those are guaranteed to pass a non-null object.

Whilst we're in here, remove a redundant local variable in `shortMessage()` and merge the `StringBuilder` calls onto a single line.

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/CompilationFailureException.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/CompilationFailureException.java
@@ -73,17 +73,11 @@ public class CompilationFailureException
      */
     public static String shortMessage( List<CompilerMessage> messages )
     {
-        StringBuilder sb = new StringBuilder();
-
-        sb.append( "Compilation failure" );
+        StringBuilder sb = new StringBuilder( "Compilation failure" );
 
         if ( messages.size() == 1 )
         {
-            sb.append( LS );
-
-            CompilerMessage compilerError = messages.get( 0 );
-
-            sb.append( compilerError ).append( LS );
+            sb.append( LS ).append( messages.get( 0 ) ).append( LS );
         }
 
         return sb.toString();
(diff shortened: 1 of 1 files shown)


rhowe:
Should this change have a Jira ticket?

src/main/java/org/apache/maven/plugin/compiler/CompilationFailureException.java
```diff
@@ -43,13 +43,11 @@ public static String longMessage( List<CompilerMessage> messages )
     {
         StringBuilder sb = new StringBuilder();
 
-        if ( messages != null )
+        for ( CompilerMessage compilerError : messages )
```
rfscholte: Let's use this opportunity to improve javadoc as well and to make clear that messages cannot be `null`.

rhowe: Sounds good to me - how's it looking now?

rhowe: Should these methods be `private static` or just inlined into the constructor? They have no other callers

rhowe: Oh, it's not possible to inline them because of the `super()` call

rhowe:
I'm afraid the tests aren't working for me locally - `master` fails in the same way as this branch using javac 11.0.4 from Debian unstable's openjdk package

rhowe:
Tests passed for me on an OS X device

rhowe:
@rfscholte any love for this little cleanup?

rfscholte:
Jenkins was happy too, so PR is merged. Thanks for the patch!

PR #21, opened 2019-05-26
Cleanup debug log handling

rhowe:
Make better use of `isDebugEnabled()`. See individual commits for details.

The diff looks better if you ignore whitespace changes.

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
@@ -963,7 +963,7 @@ public abstract class AbstractCompilerMojo
                 }
 
                 String[] cl = compiler.createCommandLine( compilerConfiguration );
-                if ( getLog().isDebugEnabled() && cl != null && cl.length > 0 )
+                if ( cl != null && cl.length > 0 )
                 {
                     StringBuilder sb = new StringBuilder();
                     sb.append( cl[0] );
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -960,7 +963,7 @@ else if ( CompilerConfiguration.CompilerReuseStrategy.ReuseSame.getStrategy().eq
                 }
 
                 String[] cl = compiler.createCommandLine( compilerConfiguration );
-                if ( getLog().isDebugEnabled() && cl != null && cl.length > 0 )
+                if ( cl != null && cl.length > 0 )
```
rfscholte: I would prefer to keep the `getLog().isDebugEnabled()` here, otherwise a lot a String concatenation is done for showing the commandline, which is only shown during debug

rhowe: We've already performed this check here: https://github.com/apache/maven-compiler-plugin/pull/21/files#diff-d4e94a0220ac4630c8e2ca12df1b46bcR929

rfscholte: Ah, good catch!

rhowe:
Should this change have a Jira ticket?

rfscholte:
I've merged it, I don't think we'll make anybody happy when such tiny improvements end up in the release notes.

PR #31, opened 2020-04-26
Prevent possible NPEx

pzygielo:
[LGTM finding](https://lgtm.com/projects/g/apache/maven-compiler-plugin/snapshot/6ae46d1998fadaea491e445edb966bbb44addc6b/files/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java#x13d75285f448c7a2:1)


rfscholte:
We prefer a style where Collections and Maps are never `null`, so you can loop without extra null-check.

pzygielo:
Is it safe to remove null-check in https://github.com/apache/maven-compiler-plugin/blob/0d1e77325d8dfb55f1a12068bbcc4be304c33b85/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java#L1347 then?

pzygielo:
Replaced with #32

PR #36, opened 2020-12-17
[MCOMPILER-447] Unable to compile modularized test for a multi-version jar that is not a java module by default

ccorsi:
The following changes resolves two different issues.  The first is the case that
the we are creating a multi-version jar file that by default is not modularized.
If tests are created that will be testing the modularized part of the multi-version
jar file.  Then the build was generating an UnsupportedOperationException because
it assumed that we were using a non-modularized jar file with a modularized test.

The second issue was related to the fact that the test are using dependency jars
that are not modularized.  The generated javac command was only producing the
required module path and nulled out the classpath.  This was causing the building
of the test sources to fail.  The was caused because those non-modularized jar
files were not part of the classpath.

The current set of changes solves the first issue by determining if the project
jar file is a multi-version jar file.  If it is, it will then set the main module
descriptor to the latest version less than or equal to the defined release
attribute.

It solves the second issue by adding all non-modularized jar files to the
class path.  It does not remove them from the defined module path to be
consistent with how it was working.

Finally an integration test was created that reproduced the above issues.

       pulling in other changes.
 - [x] Format the pull request title like `[MPH-XXX] - Fixes bug in ApproximateQuantiles`,
       commit message.

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/TestCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/TestCompilerMojo.java
@@ -27,7 +27,6 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -235,17 +234,16 @@ public class TestCompilerMojo
             int version = Integer.valueOf( release );
             while ( version >= 9 )
             {
-                String releaseOutputDirectory = String.format( "%s%sMETA-INF%sversions%s%d",
-                        getProject().getBuild().getOutputDirectory(), File.separator, File.separator,
-                        File.separator, version );
+                File metaInfoDir = new File( getProject().getBuild().getOutputDirectory(), "META-INF/versions/"
+                        + version );
 
-                File file = new File ( releaseOutputDirectory, "module-info.class" );
+                File file = new File ( metaInfoDir, "module-info.class" );
 
                 if ( file.exists() )
                 {
                     // @todo how to add the releaseOutputDirectory automatically?
                     // add releaseOutputDirectory to the beginning of the test classpath.
-                    testPath.add( 0, releaseOutputDirectory );
+                    testPath.add( 0, metaInfoDir.getAbsolutePath() );
                     mainModuleDescriptorClassFile = file;
                     if ( getLog().isDebugEnabled() )
                     {
@@ -366,12 +364,11 @@ public class TestCompilerMojo
                         {
                             if ( getLog().isDebugEnabled() )
                             {
-                                getLog().debug( "Add non-module jar dependency " + path
-                                        + " to class path." );
+                                getLog().debug( "Add non-module jar dependency " + path + " to class path." );
                             }
                             if ( classpathElements ==  null )
                             {
-                                classpathElements = new ArrayList<String>();
+                                classpathElements = new ArrayList<>();
                             }
                             // add the path to the classpathElements list.
                             classpathElements.add( path );
--- a/src/main/java/org/apache/maven/plugin/compiler/TestCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/TestCompilerMojo.java
@@ -351,35 +351,9 @@ public class TestCompilerMojo
         {
             modulepathElements = testPath;
             classpathElements = null;
-            for ( String path : modulepathElements )
-            {
-                File file = new File( path );
-                if ( file.exists() && ! file.isDirectory() && file.getName().endsWith( ".jar" ) )
-                {
-                    // check if the jar file contains the module-info.class
-                    try
-                    {
-                        JarFile jarFile = new JarFile( file );
-                        if ( jarFile.getJarEntry( "module-info.class" ) == null )
-                        {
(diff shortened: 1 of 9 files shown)


src/main/java/org/apache/maven/plugin/compiler/TestCompilerMojo.java
```diff
@@ -27,11 +27,13 @@
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.Iterator;
```
bmarwell, line 30: Please fix this breaking issue: 
`Error:  src/main/java/org/apache/maven/plugin/compiler/TestCompilerMojo.java:[30,8] (imports) UnusedImports: Unused import - java.util.Iterator.`

src/main/java/org/apache/maven/plugin/compiler/TestCompilerMojo.java
```diff
@@ -227,6 +229,37 @@ protected void preparePaths( Set<File> sourceFiles )
...
 
+        // It is possible that we are building a multi-version jar file in which the main classes are not module based.
+        if ( ! mainModuleDescriptorClassFile.exists() && release != null )
+        {
+            int version = Integer.valueOf( release );
+            while ( version >= 9 )
+            {
+                String releaseOutputDirectory = String.format( "%s%sMETA-INF%sversions%s%d",
+                        getProject().getBuild().getOutputDirectory(), File.separator, File.separator,
+                        File.separator, version );
```
bmarwell, line 240: Please use `File metaInfDir = new File(getProject().getBuild().getOutputDirectory(), "META-INF/versions/" + version)`. It will do the conversion  to system-specific separators for you.

src/main/java/org/apache/maven/plugin/compiler/TestCompilerMojo.java
```diff
@@ -319,7 +352,42 @@ else if ( Double.valueOf( getTarget() ) < Double.valueOf( MODULE_INFO_TARGET ) )
...
+                                getLog().debug( "Add non-module jar dependency " + path
+                                        + " to class path." );
+                            }
+                            if ( classpathElements ==  null )
+                            {
+                                classpathElements = new ArrayList<String>();
+                            }
+                            // add the path to the classpathElements list.
+                            classpathElements.add( path );
+                        }
```
bmarwell, line 378: Can you refactor the code a bit? Too many levels of indentation.

src/main/java/org/apache/maven/plugin/compiler/TestCompilerMojo.java
```diff
@@ -319,7 +352,42 @@ else if ( Double.valueOf( getTarget() ) < Double.valueOf( MODULE_INFO_TARGET ) )
...
+                        if ( jarFile.getJarEntry( "module-info.class" ) == null )
+                        {
+                            if ( getLog().isDebugEnabled() )
+                            {
+                                getLog().debug( "Add non-module jar dependency " + path
+                                        + " to class path." );
+                            }
+                            if ( classpathElements ==  null )
+                            {
+                                classpathElements = new ArrayList<String>();
```
bmarwell, line 374: Please use the diamond operator available in Java 7.

bmarwell (changes requested):

src/main/java/org/apache/maven/plugin/compiler/TestCompilerMojo.java
ccorsi: Done.

src/main/java/org/apache/maven/plugin/compiler/TestCompilerMojo.java
ccorsi: Done

src/main/java/org/apache/maven/plugin/compiler/TestCompilerMojo.java
ccorsi: Not sure exactly how you want me to refactor this code.  The build checkstyle definition forced me into this code style.  I did combine the debug statement within a single line.  The only other option is to extract this set of code and create a private method.  That would reduce the indentation.  Is this what you were thinking?

bmarwell: Yes that is what I had in mind. A small comment why stuff happens there would be helpful as well. I would start a little above in the "try" statement with the comment: `                    // check if the jar file contains the module-info.class.`.
I think the whole try statement would make a nice method. A simple comment why the jars are explicitly added to the classpath would absolutely suffice.

Also: 
`new ArrayList<String>();` -- you can just use the diamond operator: `new ArrayList<>();`. We are on Java 7, not 5. ;-) 

Thanks!

ccorsi: I went a step further, I ended up extracting the for loop into a separate method.

bmarwell: Well, now you have the same amount of indentation in a loop. But I will not complain anymore about indentation unless another committer does. ;-)
The comment in the new method is helpful. Nit: You could have added it as Javadoc while you were at it.

I will check it out locally in a few days to inspect the changes further.

rfscholte:
MCOMPILER-447 is closed as `Won't fix`

jira-importer:
Resolve #662

PR #42, opened 2021-03-13
Grammar on 'Compiling Sources Using A Different JDK' page

mthmulders:
@elharo You're a native speaker, could you please verify?


elharo (approved):

mthmulders:
Thank you!

PR #43, opened 2021-04-03
[MCOMPILER-417] - Document compileSourceRoots parameters

pzygielo:
This fixes [MCOMPILER-417](https://issues.apache.org/jira/browse/MCOMPILER-417) only in part of documenting _compileSourceRoots_ parameters in site page (and in `:help` goal).

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/CompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/CompilerMojo.java
@@ -65,7 +65,7 @@ public class CompilerMojo
     /**
      * The source directories containing the sources to be compiled.
      */
-    @Parameter( defaultValue = "${project.compileSourceRoots}", readonly = true, required = true )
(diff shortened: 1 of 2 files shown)


olamy (approved):

rfscholte:
I have a little bit of doubts about this. It is required to do create the MultiRelease Jars, but it feels a bit hacky. 
Normally you would use https://www.mojohaus.org/build-helper-maven-plugin/add-source-mojo.html to do this, because than it will it is a source folder for all plugins.
If we make this readable, we should mention that this is exclusive for MR Jars and include a link to the build-helper-maven-plugin.

olamy (changes requested):
@rfscholte good point

pzygielo:
Thanks for checking. Also agree with Robert's comment. Closing for now.

jira-importer:
Resolve #638

PR #88, opened 2022-02-04
[MCOMPILER-205] Add a boolean to generate missing package-info classes by default

gnodet:
https://issues.apache.org/jira/browse/MCOMPILER-205

This PR provides an option (turned on by default) to generate class files for missing `package-info` classes.  This happens when the `package-info.java` source file only contains a javadoc and no annotations on the package.
This can be turned on some JDK using `-Xpkginfo:always`, but that may not work on all JDKs.

This option is turned on by default because the generated class files are valid, so they should not cause any problem, and the current behavior means that incremental compilation is broken, which is a really annoying problem and the root cause is quite difficult to analyse.


src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -1286,6 +1315,72 @@ else if ( message.getKind() == CompilerMessage.Kind.WARNING
...
+            String path = source.toString();
+            if ( path.endsWith( File.separator + "package-info.java" ) )
+            {
+                for ( String root : getCompileSourceRoots() )
+                {
+                    root = root + File.separator;
+                    if ( path.startsWith( root ) )
+                    {
+                        String rel = path.substring( root.length() );
+                        Set<File> files = sourceMapping.getTargetFiles( getOutputDirectory(), rel );
```
jglick, line 1334: #95

joakime:
This change in MCOMPILER has started to cause problems in various tooling that scans JARs for bytecode scanning reasons.

Those users on limited JVMs (android, java 8 compact builds, embedded systems, etc) are also being impacted by this being true.

rmannibucau:
+1 it is also true on plain JVM and EE containers where it has side effects (even if some can be silently recovered) so agree default should change to be backward compatible in next release.

DarioViva42:
> so they should not cause any problem

This is a bold statement IMHO. For everyone using ArchUnit and counting on `package-info.java` files not to be compiled (as this was the default behaviour prior to 3.10.0) you can use the import-option `ImportOption.DoNotIncludePackageInfos.class` to exclude `package-info.class` files.

PR #101, opened 2022-03-12
[MCOMPILER-347] Set Xcludes in config passed to actual compiler

pzygielo:
Fixes [MCOMPILER-347: Includes and excludes not passed into CompilerConfiguration](https://issues.apache.org/jira/browse/MCOMPILER-347)

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/CompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/CompilerMojo.java
@@ -204,6 +204,18 @@ public class CompilerMojo
     }
 
     @Override
+    protected Set<String> getIncludes()
+    {
+        return includes;
+    }
+
+    @Override
+    protected Set<String> getExcludes()
+    {
+        return excludes;
+    }
+
+    @Override
     protected void preparePaths( Set<File> sourceFiles )
     {
         //assert compilePath != null;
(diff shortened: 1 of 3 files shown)


src/main/java/org/apache/maven/plugin/compiler/CompilerMojo.java
```diff
@@ -203,6 +203,18 @@ public void execute()
         }
     }
 
+    @Override
+    final Set<String> getIncludes()
```
olamy, line 207: the usual pattern in the class is to make this open with `protected` and not `final`

src/main/java/org/apache/maven/plugin/compiler/TestCompilerMojo.java
```diff
@@ -497,4 +497,16 @@ protected boolean isTestCompile()
         return true;
     }
 
+    @Override
+    final Set<String> getIncludes()
```
olamy, line 501: same `protected` and not `final`

olamy (changes requested):

src/main/java/org/apache/maven/plugin/compiler/CompilerMojo.java
pzygielo: pattern followed

PR #126, opened 2022-05-19
test build with 18 

olamy:

       pulling in other changes.
       commit message.

### DIFF
--- a/.github/workflows/maven-verify.yml
+++ b/.github/workflows/maven-verify.yml
@@ -24,8 +24,9 @@ on:
 jobs:
   build:
     name: Verify
-    uses: apache/maven-gh-actions-shared/.github/workflows/maven-verify.yml@v2
+    uses: apache/maven-gh-actions-shared/.github/workflows/maven-verify.yml@v3
     with:
       jdk-distribution-matrix: '[ "temurin", "zulu", "microsoft", "liberica", "adopt-openj9" ]'
       jdk-matrix: '[ "8", "11", "17", "18", "19-ea" ]'
       matrix-exclude: '[{ "jdk": "8", "distribution": "microsoft"}]'
+      maven-matrix: '[ "3.2.5", "3.6.3", "3.8.6" ]'  # Maven versions matrix for verify builds
--- a/.github/workflows/maven-verify.yml
+++ b/.github/workflows/maven-verify.yml
@@ -29,5 +29,5 @@ jobs:
       maven-args: "-Dinvoker.parallelThreads=2"
       jdk-distribution-matrix: '[ "temurin", "zulu", "microsoft", "liberica", "adopt-openj9" ]'
       jdk-matrix: '[ "8", "11", "17", "18", "19-ea" ]'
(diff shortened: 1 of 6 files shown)


.github/workflows/maven-verify.yml
```diff
@@ -27,4 +27,5 @@ jobs:
     uses: apache/maven-gh-actions-shared/.github/workflows/maven-verify.yml@v2
     with:
       jdk-distribution-matrix: '[ "temurin", "zulu", "microsoft", "liberica", "adopt-openj9" ]'
```
mthmulders, line 29: I know it's not part of this change, but do we really expect big differences between let's say Liberia, Zulu, Microsoft and Temurin? I do see why OpenJ9 is on the list, but all the others are regular Hotspot JVM's, aren't they?

olamy: probably not :) but as I already got differences (for javadoc) between distributions

slachiewicz (approved):

.github/workflows/maven-verify.yml
olamy: @mthmulders what would be your suggestion to limit to which?

mthmulders: Usually, I'd say "a Hotspot VM and an OpenJ9 VM". For OpenJ9, we don't have my choices; for Hotspot, I guess Temurin or maybe Zulu?

But if you're saying there are differences (e.g. in generated Javadoc) and they matter for the m-c-p, then we should include them.

.github/workflows/maven-verify.yml
```diff
@@ -24,7 +24,10 @@ on:
 jobs:
   build:
     name: Verify
-    uses: apache/maven-gh-actions-shared/.github/workflows/maven-verify.yml@v2
+    uses: apache/maven-gh-actions-shared/.github/workflows/maven-verify.yml@v3
     with:
-      jdk-distribution-matrix: '[ "temurin", "zulu", "microsoft", "liberica", "adopt-openj9" ]'
-      matrix-exclude: '[{ "jdk": "8", "distribution": "microsoft"}]'
+      maven-args: "-Dinvoker.parallelThreads=2"
```
slawekjaranowski, line 29: Why not in project configuration?
I like values as 1C, 2C, ...

olamy: because need some good environment and might not suit everyone.

olamy: btw doesn't work (e.g concurrent access failure https://github.com/apache/maven-compiler-plugin/runs/7444447535?check_suite_focus=true)

slawekjaranowski: looks like problem in Maven itself

```
Caused by: java.io.FileNotFoundException: .../local-repo/org/sonatype/plexus/plexus-build-api/0.0.4/plexus-build-api-0.0.4.pom.part (No such file or directory)
``

olamy: yup concurrent download of artifact

PR #129, opened 2022-05-30
[MCOMPILER-499] Display recompilation causes

ledoyen:

       pulling in other changes.
       commit message.

// TODO

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
@@ -60,6 +60,8 @@ import org.apache.maven.shared.incremental.IncrementalBuildHelper;
 import org.apache.maven.shared.incremental.IncrementalBuildHelperRequest;
 import org.apache.maven.shared.utils.ReaderFactory;
 import org.apache.maven.shared.utils.StringUtils;
+import org.apache.maven.shared.utils.io.DirectoryScanResult;
+import org.apache.maven.shared.utils.io.DirectoryScanner;
 import org.apache.maven.shared.utils.io.FileUtils;
 import org.apache.maven.shared.utils.logging.MessageBuilder;
 import org.apache.maven.shared.utils.logging.MessageUtils;
@@ -99,6 +101,8 @@ public abstract class AbstractCompilerMojo
     extends AbstractMojo
 {
     protected static final String PS = System.getProperty( "path.separator" );
+
+    private static final String INPUT_FILES_LST_FILENAME = "inputFiles.lst";
 
     static final String DEFAULT_SOURCE = "1.7";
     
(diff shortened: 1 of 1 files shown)


olamy:
sounds like a good idea!
But I wonder if it could be better to have this not displayed per default (as it can be very verbose)
maybe a flag such `maven.compiler.showRecompilationCause`?

ledoyen:
Notice that it is a draft, as I did not take the time to open a Jira ticket, clean the code and write tests.
However, I used the custom version on a big multi-module project to troubleshoot unwanted re-compilation and it works as intended.

That being said, to answer your concern, I introduced the flag `maven.compiler.showCompilationChanges` (defaulted to `false`) to avoid the verbose display when unwanted, and to keep the current behavior by default.

olamy:
@ledoyen thanks! and sorry for the gh actions failure but oh we do not pay for support and don't know what is the issue :)

olamy:
issue created as https://issues.apache.org/jira/browse/MCOMPILER-499

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -1515,7 +1541,14 @@ private boolean isSourceChanged( CompilerConfiguration compilerConfiguration, Co
         {
             for ( File f : staleSources )
             {
-                getLog().debug( "Stale source detected: " + f.getAbsolutePath() );
+                if ( showCompilationChanges )
```
olamy, line 1544: we cannot see it in this GH ui diff but we go in this path only `if ( getLog().isDebugEnabled() )` this must be changed to something such 
`if ( !staleSources.isEmpty() && (getLog().isDebugEnabled() ||  showCompilationChanges))`
I guess the idea is to display those files when using `-Dmaven.compiler.showCompilationChanges=true` but without using full debug with `-X`

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -1909,6 +1949,52 @@ private String getMavenCompilerPluginVersion()
...
+            {
+                oldInputFiles = FileUtils.fileReadArray( mojoConfigFile );
+            }
+            catch ( IOException e )
+            {
+                throw new MojoExecutionException( "Error reading old mojo status " + mojoConfigFile, e );
+            }
+        }
+
+        String[] inputFileNames = new String[ inputFiles.size() ];
```
olamy, line 1972:
```suggestion
String[] inputFileNames = inputFiles.stream().map(file -> file.getAbsolutePath()).toArray(String[]::new);
```

olamy (changes requested):
not sure if you want to have a look at my proposed changes or move forward with this PR.
otherwise I will do it.
thanks

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
olamy: simply replace the block with a single line. yup even Maven can use modern syntax 😜

PR #159, opened 2022-12-09
[MCOMPILER-501] compileSourceRoots parameter should be writable 

gnodet:
JIRA: https://issues.apache.org/jira/browse/MCOMPILER-501

When using multi-release JARs, the way is to configure the plugin, see https://github.com/apache/maven-compiler-plugin/blob/master/src/it/multirelease-patterns/singleproject-runtime/pom.xml#L92-L111

This causes a warning because the parameter is set as read-only.

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/CompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/CompilerMojo.java
@@ -65,7 +65,7 @@ public class CompilerMojo
     /**
      * The source directories containing the sources to be compiled.
      */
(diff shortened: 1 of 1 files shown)


slawekjaranowski (approved):
Please link with https://issues.apache.org/jira/browse/MCOMPILER-501

cstamas (approved):

olamy (approved):

slawekjaranowski:
@norrisjeremy it will be in a few days https://lists.apache.org/thread/yplrmgqzfwrr7j2ds19lf675596llp54

PR #163, opened 2022-12-16
[MCOMPILER-517] take into account last compilation time to not recompile even if user has been to install phase

olamy:
https://issues.apache.org/jira/browse/MCOMPILER-517

need this https://github.com/apache/maven-shared-incremental/pull/18

Signed-off-by: Olivier Lamy <olamy@apache.org>

       pulling in other changes.
       commit message.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -142,7 +142,7 @@ under the License.
     <dependency>
       <groupId>org.apache.maven.shared</groupId>
       <artifactId>maven-shared-incremental</artifactId>
-      <version>1.1</version>
+      <version>2.0.0-SNAPSHOT</version>
       <exclusions>
         <exclusion>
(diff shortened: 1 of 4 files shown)


eolivelli (commented):
Great stuff!

gnodet (changes requested):
I'm -1 on such a change.  This assumes the compiled classes are the same than the one in the packaged artifact, which I think is wrong.  The shade plugin for example can post-process classes, so storing the compilation time is a bad idea.  
When running multiple `mvn install`, I think the goal is to expect no artifact to be changed at all.  This will  speed up the build much more.

olamy:
yup was a bad idea. anyway my main issue is fixed with the build cache extension

jira-importer:
Resolve #729

PR #167, opened 2023-01-12
[MCOMPILER-520] Upgrade parent to version 39 + reformat code

psiroky:
* I reformatted also the IT projects so that the code there is somewhat consistent with the main code base. That makes the PR a bit bigger, so I could possibly split that up if needed

Following this checklist to help us incorporate your contribution quickly and easily:

       pulling in other changes.
       commit message.

If your pull request is about ~20 lines of code you don't need to sign an [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf) if you are unsure please ask on the developers list.


pom.xml
```diff
@@ -208,6 +193,19 @@ under the License.
...
+      <releases>
+        <enabled>false</enabled>
+      </releases>
+      <snapshots>
+        <enabled>true</enabled>
+      </snapshots>
+      <id>plexus.snapshots</id>
+      <url>https://oss.sonatype.org/content/repositories/plexus-snapshots</url>
+    </repository>
+  </repositories>
```
psiroky, line 207: I am wondering if this correct? Should we have another repository directly hardcoded in the POM?

I know this has been used to get SNAPSHOTS in the past, but currently the plexus-compiler is a release version, and I assume it should stay that way.

psiroky:
This actually does loads of changes and is likely impossible to reasonably review.

Probably better if one of the trusted core committers does the reformatting - at least for the production code, so the review would not really be needed.

I am keeping the branch around if anyone would like to reuse it.

PR #151, opened 2022-09-26
Move away from legacy maven-compat

cstamas:
And switch to resolver to resolver annotation processors with dependencies.

PR #169, opened 2023-01-19
[MCOMPILER-522] Use maven-resolver to resolve 'annotationProcessorPaths' dependencies

psiroky:
* heavily based on work by @cstamas (PR #151)

 * added IT to verify the build fails if one the dependencies cannot be resolved (e.g. does not exist)

       pulling in other changes.
       commit message.

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
@@ -40,14 +40,8 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
 
-import org.apache.maven.artifact.Artifact;
-import org.apache.maven.artifact.DefaultArtifact;
 import org.apache.maven.artifact.handler.ArtifactHandler;
 import org.apache.maven.artifact.handler.manager.ArtifactHandlerManager;
-import org.apache.maven.artifact.resolver.ArtifactResolutionRequest;
-import org.apache.maven.artifact.resolver.ArtifactResolutionResult;
-import org.apache.maven.artifact.resolver.ResolutionErrorHandler;
-import org.apache.maven.artifact.versioning.VersionRange;
 import org.apache.maven.execution.MavenExecutionRequest;
 import org.apache.maven.execution.MavenSession;
 import org.apache.maven.plugin.AbstractMojo;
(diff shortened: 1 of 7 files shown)


psiroky:
@ajarmoniuk would you have some time to review? (since you did similar changes in other plugins)

ajarmoniuk:
No problem, I'll do that in the evening

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -1814,29 +1810,26 @@ private List<String> resolveProcessorPathEntries()
...
+                DependencyResult dependencyResult = repositorySystem.resolveDependencies(
+                        session.getRepositorySession(), dependencyRequest );
+
+                for ( ArtifactResult resolved : dependencyResult.getArtifactResults() )
                 {
-                    elements.add( resolved.getFile().getAbsolutePath() );
+                    if ( resolved.getArtifact() != null && resolved.getArtifact().getFile() != null )
+                    {
+                        elements.add( resolved.getArtifact().getFile().getAbsolutePath() );
+                    }
```
psiroky, line 1832: Not sure about this. I assume the artifact, nor the path should be null if the resolving is successful. However, if we explicitly check, maybe we should throw an exception in the `else` branch? Otherwise we would just not include on of the dependencies without any kind of message to the user.

ajarmoniuk: Correct. In addition, result.getExceptions() will contain the list of all gathered exceptions (if there are more than one).

ajarmoniuk (approved):

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
ajarmoniuk: If that's the case then it looks like a bug. Did you actually reproduce that case?

cstamas: Remove null check from here, it was my mistake. If nothing thrown, all resolved just fine: https://github.com/apache/maven-resolver/blob/master/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultArtifactResolver.java#L486-L509

ajarmoniuk: Just use Objects.requireNonNull

psiroky: I removed the `null` check and marking this as resolved based on the above discussion.

cstamas (approved):

slachiewicz (approved):

PR #170, opened 2023-01-22
[MCOMPILER-503] Resolve all annotation processor dependencies together

psiroky:
* resolving the anootation processor paths in multiple requests can easily lead to duplicated dependencies, e.g. if two processors depend on the same transitive dependency. Ultimately, both versions would end-up on processorpath, but of course only one of them would be used (likely dependeing on the order of elements of processorpath)

 * the previous behavior also did not match the general rules of dependency resolution used by Maven itself and other plugins

 * this may potentially be a breaking change for certain corner cases, if for example some projects mistakenly rely on the order of elemenets in processorpath

       pulling in other changes.
       commit message.

 - [x] I hereby declare this contribution to be licensed under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)


psiroky:
I would like to create a test for this, but I am not sure which approach to take. Unit testing this seems somewhat not-relevant, since all the code does is calling maven-resolver, which does the actual heavy lifting.

Creating an integration test project is pretty easy, but how do I for example verify, that duplicated dependencies are not on processor path? Parsing he debug (-X) Maven output (where the modulepath elements are listed) seems pretty brittle. Is there a better way to do this?

And one more question related to ITs, and the dependencies they use: is it OK to depend on a third party dep (say junit) in the IT projects? Those deps would need to be downloaded from Maven central at runtime (test time). I noticed there are ITs doing that, but I would like to clarify if that's just an oversight, or a proper way of writing the tests (since `maven-integration-testing` seems to be a bit more strict on how the dependencies can be used by tests).

cstamas:
One big fat warning about this approach, or in general for "resolving": if you look at old code carefully, you will see it did `new CollectRequest( new Dependency( artifact, JavaScopes.RUNTIME ), project.getRemoteProjectRepositories() )` (pay attention to `CollectRequest` ctor), it used current artifact _as root_ and resolved it's dependencies. What new code does is pushing all the enlisted artifacts as "direct dependencies" of some (imaginary) root (again, according to Javadoc of changed ctor for `CollectRequest`). This has slightly different meaning and semantics, especially with optional dependencies, see https://github.com/apache/maven-resolver/blob/master/maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/selector/OptionalDependencySelector.java#L52-L56

The change in semantics comes from fact that new code will "push" all things +1 regarding depth (as they each are not root anymore, but are direct siblings on some imaginary root).

In this very case (compiler annotation processors) this is completely OK, but when you try to "mimic" what Maven does not result in same outcome, for example when Maven resolves a plugin, it is used as root, and hence, this approach in this PR _cannot_ be applied to bootstrap IT here (see https://github.com/apache/maven-integration-testing/pull/236 and example mentioned in it).

cstamas (approved):

psiroky:
@cstamas many thanks for the review (and the details around dependency resolution)! I am wondering - what else do we need to get this one merged?

cstamas:
we need Slawek 😄 @slachiewicz

psiroky:
> we need Slawek

I see :laughing:. I thought you have the merge permission as well, sorry about the fuzz.

cstamas:
I do, but two pairs of eyes are better than m one

slawekjaranowski:
I don't see where plugin from annotation-verify is executed.

We can put such plugin in separate setup-xxx test with install goal and next use plugin in your IT test.

By the way ... configuration for invoker in project is to fix - I will take care about it.

psiroky:
The plugin is used in the `annotation-user` project: https://github.com/apache/maven-compiler-plugin/pull/170/files#diff-82b7fbab158b18d75e40832a9a3edfee28aba32e67314a155f1bbc923f9b84e9R58. Just to make sure the annotation processor was actually executed and created those dummy files.

I agree we should reuse the plugin, instead of copying it (it is now in three tests I think). Do you want me to do that in this PR, or in the next one?

pom.xml
```diff
@@ -249,13 +249,8 @@ under the License.
...
                       -->
                     <setupIncludes>
-                      <setupInclude>setup_jar_module/pom.xml</setupInclude>
-                      <setupInclude>setup_jar_automodule/pom.xml</setupInclude>
-                      <setupInclude>setup_x/pom.xml</setupInclude>
+                      <setupInclude>setup*/pom.xml</setupInclude>
                     </setupIncludes>
-                    <setupExcludes>
-                      <setupExclude>setup_x/**</setupExclude>
-                    </setupExcludes>
```
psiroky, line 258: Excludes are not supported by the invoker plugin, so this configuration was a noop

pom.xml
```diff
@@ -249,13 +249,8 @@ under the License.
                       ! https://issues.apache.org/jira/browse/MINVOKER-174
                       -->
                     <setupIncludes>
-                      <setupInclude>setup_jar_module/pom.xml</setupInclude>
-                      <setupInclude>setup_jar_automodule/pom.xml</setupInclude>
-                      <setupInclude>setup_x/pom.xml</setupInclude>
+                      <setupInclude>setup*/pom.xml</setupInclude>
```
psiroky, line 252: This makes sure that any new `setup` projects following the naming convention will be automatically picked up. If there is any potential concern with this approach, I can revert to the explicit list.

src/it/setup_annotation-verify-plugin/pom.xml
```diff
@@ -1,123 +1,119 @@
...
+          <version>@mavenPluginPluginVersion@</version>
+          <configuration>
+            <skipErrorNoDescriptorsFound>true</skipErrorNoDescriptorsFound>
+          </configuration>
+          <executions>
+            <execution>
+              <id>mojo-descriptor</id>
+              <phase>process-classes</phase>
+              <goals>
+                <goal>descriptor</goal>
```
slawekjaranowski, line 112: it is not needed, for `maven-plugin` packaging is bind by default to lifecycle

slawekjaranowski (approved):

src/it/setup_annotation-verify-plugin/pom.xml
psiroky: Yeah, I just copy-pasted the existing stuff. Let me fix that.

src/it/setup_annotation-verify-plugin/pom.xml
```diff
@@ -1,123 +1,119 @@
...
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.maven.plugin-tools</groupId>
+      <artifactId>maven-plugin-annotations</artifactId>
+      <version>3.3</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>maven-plugin-api</artifactId>
```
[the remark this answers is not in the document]
psiroky: Added the `provided` scope and also removed the dependency on `maven-shared-utils`, since the functionality the plugin needs is now in the standard Java 8 library.

psiroky:
@cstamas @slawekjaranowski hey, just checking in - is there anything else that needs to happen to get this one merged?

(I definitely don't want to be pushing you or anything, I am sure you have more than enough on your plate. Just trying to figure out if there are other changes needed or so.)

src/it/setup_annotation-verify-plugin/src/main/java/org.apache.maven.plugins.compiler.it/SourcePathReadGoal.java
```diff
@@ -80,13 +79,12 @@ private void assertGeneratedSourceFileFor( String sourceClass, List<String> sour
                 try
                 {
                     String[] nameParts = sourceClass.split( "\\." );
-                    String content = FileUtils.fileRead( f );
+                    String content = new String( Files.readAllBytes( f.toPath() ) );
```
Stephan202, line 82: I wonder: shouldn't this code call one of the [two-arg `String` constructors](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/String.html#%3Cinit%3E(byte%5B%5D,java.lang.String)) and pass in the charset specified by `project.build.sourceEncoding`?

Stephan202 (commented):
Thanks for resolving [MCOMPILER-503](https://issues.apache.org/jira/browse/MCOMPILER-503), @psiroky :pray:

src/it/setup_annotation-verify-plugin/src/main/java/org.apache.maven.plugins.compiler.it/SourcePathReadGoal.java
psiroky: Yes, that would be ideal. The plugin is only used by ITs in this repo, which are fully under "our" control, so the risk of relying on the platform encoding seems small. In any case, I will take a look at this in my next PR.

psiroky: The "fix" is now part of https://github.com/apache/maven-compiler-plugin/pull/173.

It could be definitely further improved, e.g. issuing a WARNing as `maven-resources-plugin` does when the encoding is not set. That being said, considering this plugin is only used in the ITs (and all of them are under our control), the solution seems fine to me (it will respect the source encoding if specified in the test project and will fail horribly if e.g. the encoding is unknown / wrong).

PR #172, opened 2023-01-29
[MCOMPILER-525] Incorrect detection of dependency change

jorsol:
This [fix a bug](https://issues.apache.org/jira/browse/MCOMPILER-525) with the incremental compilation related to the incorrect detection of the project classes as a dependency.

 - [X] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MCOMPILER) filed 
       pulling in other changes.
 - [X] Format the pull request title like `[MCOMPILER-XXX] - Fixes bug in ApproximateQuantiles`,
       commit message.


src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -902,7 +906,7 @@ else if ( CompilerConfiguration.CompilerReuseStrategy.ReuseSame.getStrategy().eq
                 {
                     String cause = idk ? "idk" : ( dependencyChanged ? "dependency"
                             : ( sourceChanged ? "source" : "input tree" ) );
-                    getLog().info( "Changes detected - recompiling the module! :" + cause );
+                    getLog().info( "Changes detected! - recompiling the module due to: " + cause );
```
pzygielo, line 909: If this message is going to change - could we drop exclamation point completely?

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -1733,30 +1740,46 @@ protected boolean isDependencyChanged()
...
         {
-            File artifactPath = new File( pathElement );
-            if ( artifactPath.isDirectory() || artifactPath.isFile() )
+            Path artifactPath = Paths.get( pathElement );
+
+            // Search files only on dependencies (other modules), not the current project.
+            if ( Files.isDirectory( artifactPath ) && !artifactPath.equals( getOutputDirectory().toPath() ) )
             {
-                if ( hasNewFile( artifactPath, buildStartTime ) )
+                try ( Stream<Path> walk = Files.walk( artifactPath ) )
```
olamy, line 1756: I need to review more this. At this may have bad performance impact on large projects with a lot of modules and large source code (e.g source files number).
this PR https://github.com/apache/maven-compiler-plugin/pull/163 use another path by storing some information rather than walking again and again the same source tree.

olamy:
@jorsol thanks for your PR. is your issue fixed if you try this PR https://github.com/apache/maven-compiler-plugin/pull/172/files#r1090053416?
because I'm a bit scared of the performance impact of some changes here

pzygielo:
I believe https://github.com/apache/maven-compiler-plugin/blob/3f95d39f166809bc1a2bb0b2fa62c190e8d2a9f7/src/it/MCOMPILER-275_separate-moduleinfo/verify.groovy#L23 should be updated as well.
It's skipped on GitHub, but it fails.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -531,12 +541,12 @@
...
     /**
      * File extensions to check timestamp for incremental build.
-     * Default contains only <code>class</code> and <code>jar</code>.
+     * Default contains only {@code class} and {@code jar}.
      *
      * @since 3.1
      */
-    @Parameter
-    private List<String> fileExtensions;
+    @Parameter( defaultValue = "class,jar" )
```
slawekjaranowski, line 548: When you add `defaultValue` the information about it will added by plugin tools, so we can remove manual description.

src/it/MCOMPILER-525/pom.xml
```diff
@@ -0,0 +1,64 @@
...
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>org.apache.maven.plugins</groupId>
+  <artifactId>MJAR-70-no-recreation</artifactId>
```
slawekjaranowski, line 23: `MJAR-70` -> `MCOMPILER-525` - what do yo think?

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -1917,7 +1965,7 @@ private DirectoryScanResult computeInputFileTreeChanges( IncrementalBuildHelper
             }
             catch ( IOException e )
             {
-                throw new MojoExecutionException( "Error reading old mojo status " + mojoConfigFile, e );
+                throw new UncheckedIOException( "Error reading old mojo status " + mojoConfigFile, e );
```
olamy, line 1968: why changing that?

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
olamy: my concern is the method `computeInputFileTreeChanges` already scanning files and here you add another source tree scanning so maybe `IncrementalBuildHelper` could be improved to store the information you need? (in a similar logic as this PR https://github.com/apache/maven-compiler-plugin/pull/163 ).
?
regarding `large project` what sort of figures? because this `large` can be relative ;)

olamy (changes requested):

gnodet:
@jorsol I have a hard time understanding the exact change that fixes the problem.  Would it be possible to have a first commit that refactors the code using the nio api and other improvements, and another commit with the actual fix for the problem ?

Note that we could also embeds [the trimmed down version](https://github.com/apache/maven-compiler-plugin/blob/7d9b2f7dbfd51d8400c332f6f50388d5a3eaf100/src/main/java/org/apache/maven/plugin/compiler/IncrementalBuildHelper.java) of the `IncrementalBuildHelper`.

gnodet (approved):

PR #173, opened 2023-01-30
[MCOMPILER-395] Allow dependency exclusions for 'annotationProcessorPaths'

psiroky:
* this should be quite straightforward to review, since the plugin only needs to "push" the exclusions down to the maven-resolver and all the "magic" happens there

       pulling in other changes.
       commit message.


src/main/java/org/apache/maven/plugin/compiler/DependencyExclusion.java
```diff
@@ -0,0 +1,114 @@
...
+    @Override
+    public int hashCode()
+    {
+        final int prime = 31;
+        int result = 1;
+        result = prime * result + ( ( artifactId == null ) ? 0 : artifactId.hashCode() );
+        result = prime * result + ( ( classifier == null ) ? 0 : classifier.hashCode() );
+        result = prime * result + ( ( groupId == null ) ? 0 : groupId.hashCode() );
+        result = prime * result + ( ( classifier == null ) ? 0 : classifier.hashCode() );
+        result = prime * result + ( ( extension == null ) ? 0 : extension.hashCode() );
```
psiroky, line 104: I would be very much in favor of using the simplified version like this
```
return Objects.hash( groupId, artifactId, classifier, extension );
```
However, this will create a new array every time the method is called, which could potentially be a performance concern (since currently the hashCode() does not allocate anything). I don't think this method will be used frequently though, since the plugin only creates an instance and then transforms it into Aether Exclusion class.

Would be great to get feedback from others on this.

psiroky:
@slawekjaranowski @cstamas hello again! Would you guys have some time to review this PR?

src/main/java/org/apache/maven/plugin/compiler/DependencyExclusion.java
slawekjaranowski: looks better

src/main/java/org/apache/maven/plugin/compiler/DependencyCoordinate.java
```diff
@@ -107,71 +123,17 @@ public boolean equals( Object obj )
...
-            if ( other.version != null )
-            {
-                return false;
-            }
-        }
-        else if ( !version.equals( other.version ) )
-        {
-            return false;
-        }
-        return true;
```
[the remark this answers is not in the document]
slawekjaranowski: 😄

slawekjaranowski (commented):
Please update documentation of `annotationProcessorPaths` parameter.

psiroky:
@slawekjaranowski thansk, Javadoc updated (I hope this is what you meant by documentation). The rendered page now looks like this:
compiler-plugin-path-exclusions

src/it/setup_annotation-verify-plugin/src/main/java/org.apache.maven.plugins.compiler.it/SourcePathReadGoal.java
```diff
@@ -38,15 +38,17 @@
...
     @Parameter
-    protected String testSourceClass;
+    private String testSourceClass;
 
     @Parameter( defaultValue = "${project}" )
-    protected MavenProject project;
+    private MavenProject project;
+
+    @Parameter( defaultValue = "${project.build.sourceEncoding}" )
+    private String encoding;
```
gnodet, line 50: Can this be moved in other PR ?

gnodet (changes requested):

src/it/setup_annotation-verify-plugin/src/main/java/org.apache.maven.plugins.compiler.it/SourcePathReadGoal.java
psiroky: Sure, let me remove that.

psiroky:
The failed test is `MCOMPILER-481-requires-static-included`, which seems to a [known issue](https://issues.apache.org/jira/browse/MCOMPILER-481?focusedCommentId=17679050&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17679050). The same test fails also on master with Maven 3.9.0. It is possibly related to surefire version 3.0.0.M8 (which is the default in Maven 3.9.0 I believe). Temporary workaround could be to pin down the surefire plugin version to some older one for that specific test case (`2.22.2` seems to be working).

gnodet (approved):

cstamas (approved):

PR #175, opened 2023-02-10
[MCOMPILER-526] Upgrade to parent 39

gnodet:
- [MCOMPILER-526] Upgrade to parent 39
- [MCOMPILER-526] Reformat

Requires https://github.com/apache/maven-compiler-plugin/pull/176

       pulling in other changes.
       commit message.


psiroky:
Yes, please! Thank you for this change. `<personalRant>` the old style was really hurting my eyes `</personalRant>`

src/it/MCOMPILER-474_recompile-dependent-when-package/dependent-module/src/main/java/Main.java
```diff
@@ -16,11 +16,10 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+package foo;
```
slawekjaranowski, line 19: Should we move file to `.../java/foo/Main.java` to be consistent with package name?

gnodet: We could, but there are tons of files to move manually for no benefit, and those are just tests, so I left it the way it is.  are you fancy doing it ?  I did not found  a trivial way to do that. 
Fwiw, the java compiler puts the compiled classes in the correct `foo/Main.class` while the groovy compile does not, so I had to move the groovy ones.

src/it/MCOMPILER-503-processorpath-duplicated-deps/annotation-user/src/main/java/mcompiler503/SimpleObject.java
```diff
@@ -1,3 +1,21 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
```
slawekjaranowski, line 2: double license header 😄

slawekjaranowski (approved):

PR #180, opened 2023-02-15
[MCOMPILER-391] Use dep mgmt when resolving annotation processors and their deps

psiroky:
* dependency managemet can be used for two slighty difference use
   cases (when it comes to annotation processors):
     -- getting the version of the top-level processor path elements
     -- passing the list of managed dependencies to maven-resolver
        when resolving the whole processorpath
   These two can be combined, so there is total of 4 valid combinations
   (some of them make more sense than the others, but generally there is
   no reason to not support all of them)

 * using dependency management when resolving transitive dependencies of
   annotation processors is something that may or may not be desired.
   For that reason, there is a new flag that explicitly enables this
   behavior, while default is to _not_ use dependency management
   (the current behavior)

 * do not be scared by the the number of new lines. 95% of that are ITs, and of that like 50% are license headers :smile: 

These are the 4 use cases I am considering as part of this PR:
1. Version of the annotation processor path is explicitly set + useDepMgmt flag = false (default) -- the current behavior, dependency management is not used at all

2. Version of the annotation processor path is explicitly set + useDepMgmt flag = true (this may make the least sense of the combinations, but I see no reason to not allow it)

3. Version of the annotation processor dep is taken from dep mgmt + useDepMgmt flag = false (dep mgmt is when resolving the whole processorpath).

4. Version of the annotation processor dep is taken from dep mgmt + useDepMgmt flag= true (dep mgmt used when resolving the whole processorpath)

Note: the whole logic around annotation processor paths is getting bit out of hand (read more complex) and I think it would make sense to extract this into a separate class. I did want to complicate this PR further though, so that would be something for next PR(s).

       pulling in other changes.
       commit message.

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
@@ -36,13 +36,19 @@ import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
+import java.util.stream.Collectors;
 
+import org.apache.maven.RepositoryUtils;
 import org.apache.maven.artifact.handler.ArtifactHandler;
 import org.apache.maven.artifact.handler.manager.ArtifactHandlerManager;
 import org.apache.maven.execution.MavenExecutionRequest;
 import org.apache.maven.execution.MavenSession;
+import org.apache.maven.model.Dependency;
+import org.apache.maven.model.DependencyManagement;
 import org.apache.maven.plugin.AbstractMojo;
 import org.apache.maven.plugin.MojoExecution;
 import org.apache.maven.plugin.MojoExecutionException;
@@ -77,11 +83,10 @@ import org.codehaus.plexus.languages.java.jpms.JavaModuleDescriptor;
 import org.codehaus.plexus.languages.java.version.JavaVersion;
 import org.eclipse.aether.RepositorySystem;
 import org.eclipse.aether.artifact.Artifact;
+import org.eclipse.aether.artifact.ArtifactTypeRegistry;
 import org.eclipse.aether.artifact.DefaultArtifact;
 import org.eclipse.aether.collection.CollectRequest;
-import org.eclipse.aether.graph.Dependency;
 import org.eclipse.aether.graph.Exclusion;
-import org.eclipse.aether.resolution.ArtifactResult;
 import org.eclipse.aether.resolution.DependencyRequest;
 import org.eclipse.aether.resolution.DependencyResult;
 import org.eclipse.aether.util.artifact.JavaScopes;
@@ -309,7 +314,7 @@ public abstract class AbstractCompilerMojo extends AbstractMojo {
      *     &lt;path&gt;
      *       &lt;groupId&gt;org.sample&lt;/groupId&gt;
      *       &lt;artifactId&gt;sample-annotation-processor&lt;/artifactId&gt;
-     *       &lt;version&gt;1.2.3&lt;/version&gt;
+     *       &lt;version&gt;1.2.3&lt;/version&gt; &lt;!-- Optional - taken from dependency management if not specified --&gt;
      *       &lt;!-- Optionally exclude transitive dependencies --&gt;
      *       &lt;exclusions&gt;
      *         &lt;exclusion&gt;
@@ -329,6 +334,22 @@ public abstract class AbstractCompilerMojo extends AbstractMojo {
      */
     @Parameter
     private List<DependencyCoordinate> annotationProcessorPaths;
+
+    /**
+     * <p>
+     * Whether to use the Maven dependency management section when resolving transitive dependencies of annotation
+     * processor paths.
+     * </p>
+     * <p>
+     * This flag does not enable / disable the ability to resolve the version of annotation processor paths
+     * from dependency management section. It only influences the resolution o transitive dependencies of those
+     * top-level paths.
+     * </p>
+     *
+     * @since 3.12.0
+     */
+    @Parameter(defaultValue = "false")
+    private boolean annotationProcessorPathsUseDepMgmt;
 
     /**
      * <p>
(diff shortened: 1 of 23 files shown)


src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -1604,22 +1606,39 @@ private List<String> resolveProcessorPathEntries() throws MojoExecutionException
...
+            throws MojoExecutionException {
         List<Dependency> dependencies = new ArrayList<>();
         for (DependencyCoordinate annotationProcessorPath : annotationProcessorPaths) {
             ArtifactHandler handler = artifactHandlerManager.getArtifactHandler(annotationProcessorPath.getType());
+            String version = annotationProcessorPath.getVersion();
+            if (version == null) {
+                version = findManagedVersion(annotationProcessorPath)
+                        .orElseThrow(
+                                () -> new MojoExecutionException("Cannot find version for " + annotationProcessorPath));
+            }
```
psiroky, line 1619: It would be ideal if this wasn't needed and we could just rely on `maven-resolver` to simply figure out the version (if it's null) from the supplied list of managed versions. However, I did not find a way to create a `CollectRequest` which would enable this. The resolving always fails with 
```
Caused by: java.lang.IllegalArgumentException: version can neither be null, empty nor blank
	at org.apache.commons.lang3.Validate.notBlank(Validate.java:454)
	at org.apache.maven.artifact.ArtifactUtils.notBlank(ArtifactUtils.java:107)
	at org.apache.maven.artifact.ArtifactUtils.key(ArtifactUtils.java:97)
	at org.apache.maven.ReactorReader.findArtifact(ReactorReader.java:101)
	at org.eclipse.aether.internal.impl.DefaultArtifactResolver.resolve(DefaultArtifactResolver.java:306)
	at org.eclipse.aether.internal.impl.DefaultArtifactResolver.resolveArtifacts(DefaultArtifactResolver.java:229)
	at org.eclipse.aether.internal.impl.DefaultArtifactResolver.resolveArtifact(DefaultArtifactResolver.java:207)
	at org.apache.maven.repository.internal.DefaultArtifactDescriptorReader.loadPom(DefaultArtifactDescriptorReader.java:240)
	at org.apache.maven.repository.internal.DefaultArtifactDescriptorReader.readArtifactDescriptor(DefaultArtifactDescriptorReader.java:171)
	at org.eclipse.aether.internal.impl.collect.DefaultDependencyCollector.resolveCachedArtifactDescriptor(DefaultDependencyCollector.java:538)
	at org.eclipse.aether.internal.impl.collect.DefaultDependencyCollector.getArtifactDescriptorResult(DefaultDependencyCollector.java:523)
	at org.eclipse.aether.internal.impl.collect.DefaultDependencyCollector.processDependency(DefaultDependencyCollector.java:410)
	at org.eclipse.aether.internal.impl.collect.DefaultDependencyCollector.processDependency(DefaultDependencyCollector.java:362)
	at org.eclipse.aether.internal.impl.collect.DefaultDependencyCollector.process(DefaultDependencyCollector.java:349)
	at org.eclipse.aether.internal.impl.collect.DefaultDependencyCollector.collectDependencies(DefaultDependencyCollector.java:254)
	at org.eclipse.aether.internal.impl.DefaultRepositorySystem.resolveDependencies(DefaultRepositorySystem.java:309)
	at org.apache.maven.plugin.compiler.AbstractCompilerMojo.resolveProcessorPathEntries(AbstractCompilerMojo.java:1595)
```

I may have missed something, so any suggestions are more than welcome.

cstamas: Well, even in case of dependencies, it is maven that does this (by calculating effective POM), not resolver :smile:

psiroky: Sorry, I am not sure I quite got that.

I was thinking that resolver could check if the version is null and if so, then try to find the version from the supplied managed deps (somewhere inside `org.eclipse.aether.internal.impl.DefaultRepositorySystem.resolveDependencies`). But maybe that's not really what resolver should be doing anyway. And I assume that even if this was added, the plugin could not rely on this (yet), because the older Maven versions (with older resolvers) would not support that.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -330,6 +335,22 @@ public abstract class AbstractCompilerMojo extends AbstractMojo {
...
+     * </p>
+     * <p>
+     * This flag does not enable / disable the ability to get the version of annotation processor paths
+     * (the top-level paths) from dependency management section. It only influences the resolution of
+     * transitive dependencies of those top-level paths.
+     * </p>
+     *
+     * @since 3.12.0
+     */
+    @Parameter(defaultValue = "false")
```
cstamas, line 351: :beer: for this: new feature, if alters behavior, should be disabled by def

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -330,6 +335,22 @@ public abstract class AbstractCompilerMojo extends AbstractMojo {
...
+    private boolean annotationProcessorPathsUseDepMgmt;
```
psiroky, line 352: I am not the biggest fan of this long name, but I could not find a better one at this point. Reason I choose this one is that the prefix `annotationProcessorPaths` means it will be rendered/shown together with the paths option, meaning users are much more likely to see this when browsing the docs (and making the connection that is used for annotation processor paths). We could use something like `useDepMgmt` or so, but then you need to know exactly what you are looking for, since this would be rendered very far away from the `annotationProcessorPaths`. Also something like `useDepMgmt` may be too generic?

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -330,6 +335,22 @@ public abstract class AbstractCompilerMojo extends AbstractMojo {
...
+     * Whether to use the Maven dependency management section when resolving transitive dependencies of annotation
+     * processor paths.
+     * </p>
+     * <p>
+     * This flag does not enable / disable the ability to get the version of annotation processor paths
+     * (the top-level paths) from dependency management section. It only influences the resolution of
+     * transitive dependencies of those top-level paths.
+     * </p>
+     *
+     * @since 3.12.0
```
psiroky, line 349: New feature, so I think this should go into the next minor version (not patch).

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -1584,42 +1605,91 @@ private List<String> resolveProcessorPathEntries() throws MojoExecutionException
...
+    private Optional<String> findManagedVersion(
+            DependencyCoordinate dependencyCoordinate, List<Dependency> managedDependencies) {
+        return managedDependencies.stream()
+                .filter(dep -> Objects.equals(dep.getGroupId(), dependencyCoordinate.getGroupId())
+                        && Objects.equals(dep.getArtifactId(), dependencyCoordinate.getArtifactId())
+                        && Objects.equals(dep.getClassifier(), dependencyCoordinate.getClassifier())
+                        && Objects.equals(dep.getType(), dependencyCoordinate.getType()))
+                .findAny()
+                .map(org.apache.maven.model.Dependency::getVersion);
+    }
```
psiroky, line 1670: This could be somewhat problematic from performance point of view, in case the dependency management section is big (e.g. thousands of dependencies, which is not that uncommon I think -- thinking about SpringBoot or Quarkus BOMs). On the other hand, I have not found a better way at this point.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -1584,42 +1605,91 @@ private List<String> resolveProcessorPathEntries() throws MojoExecutionException
...
+    private List<org.eclipse.aether.graph.Dependency> getManagedDependenciesForAnnotationProcessorPaths() {
+        if (!annotationProcessorPathsUseDepMgmt) {
+            return Collections.emptyList();
+        }
+        List<Dependency> projectManagedDependencies = getProjectManagedDependencies();
+        ArtifactTypeRegistry artifactTypeRegistry =
+                session.getRepositorySession().getArtifactTypeRegistry();
+
+        return projectManagedDependencies.stream()
+                .map(dep -> RepositoryUtils.toDependency(dep, artifactTypeRegistry))
```
psiroky, line 1681: This is also not great, since it re-creates the dependency list for every sub-module of the Maven build, and usually the dependency management section is 99% same for all modules.

Taking example of Quarkus. It has ~500 modules using annotation processing and the dependency management section has ~2k dependencies. That 1 million of new objects created during the build. Possible not the biggest deal (since the build already creates tens of millions of other objects). Again, at this point I am not seeing a simple alternative. Suggestions more than welcome.

Maybe some caching could be done here. But caching is usually double-edge sword, so I wanted to avoid that for now.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
psiroky: It may definitely change the resulting `processorpath`, thus changing the behavior. So it should be opt-in.

psiroky:
I decided to proceed with the suggested solution. The PR now contains changes for both use cases:
 1. using dependency management to get just the top-level annotation processor path version
 2. using dependency management when resolving the transitive dependencies via maven-resolver

There is also a new flag that enables / disables the usage of dependency management for the second use case.

I added some inline comments for things that I am not 100% sure about or things that potentially need more eyes.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
psiroky: Thinking about this again, this would probably mean we could not distinguish between the 2 difference cases: version for the top-level path, and managed versions for the transitive dependencies. Marking this as resolved.

psiroky:
I ran a quick performance test using the Quarkus repo (1k modules, where 500 of them are using annotation processors) and I could not really spot any noticeable difference in the overall build time. I am sure there is more "work" being done (e.g. search those dependency management lists), but it seems it gets lost in the rest of the work being done as part of the build. I ran the following command:
```
mvn387 clean install -DskipTests -Dversion.enforcer.plugin=3.2.1 -Dversion.compiler.plugin=3.11.1-SNAPSHOT -T8
```
with three different configurations: 
 * default (no dependency management usage) - option 1) above
 * using dep mgmt to only get the top-level versions - option 3) above
 * using dep mgmt to get the top-level versions and also to resolve transitive dependencies - option 4)

In all cases I got build times ranging from 01:49min to 01:52min. No noticeable difference overall.

slawekjaranowski (approved):
Very good work.
Thanks for comments with explanation.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
psiroky: I can add a note to the docs.

Not sure about `parallelStream` (especially in combination with -T), but I could try and test to see if there is any improvement.

graemerocher:
could we kindly ask for an increase in priority on this issue. The fact that Maven is unable to correctly manage annotation processor dependencies like Gradle can is a major disadvantage. It is issues like this that imagine lead to any tool chain that relies on annotation processors moving away from Maven (see Android)

psiroky:
I can update the docs, based on the comments above. Besides that, the PR should be ready from my point of view. Please let me know if there is anything else that would need changing.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
slawekjaranowski: ok, we can go as is - if we need we can tray with `parallelStream` next time 😄

slachiewicz (approved):

famod:
Just a heads up: This implementation does not seem to respect relocations, e.g. I was still using https://mvnrepository.com/artifact/org.hibernate/hibernate-jpamodelgen in my plugin config but the Quarkus BOM contains (only) the new and proper group id `org.hibernate.orm` and not the old one ``org.hibernate.` and so the annotation processor path version lookup logic didn't find it. Took me a while to figure out.

famod:
@psiroky sure:
> Are you relying on the new dependency management

Yes, that's what I tried and this doesn't work (old groupId):
```xml
                <configuration>
                    <annotationProcessorPaths>
                        <path>
                            <groupId>org.hibernate</groupId>
                            <artifactId>hibernate-jpamodelgen</artifactId>
                        </path>
```
but this does work (new, relocated groupId):
```

```
                <configuration>
                    <annotationProcessorPaths>
                        <path>
                            <groupId>org.hibernate.orm</groupId>
                            <artifactId>hibernate-jpamodelgen</artifactId>
                        </path>
```
The Quarkus BOM I'm importing has this entry (new groupId only!):
https://github.com/quarkusio/quarkus/blob/3.6.4/bom/application/pom.xml#L5012-L5014

**Comment by psiroky** [comment]
@famod thanks. In this case, I am not quite sure this is supported from Maven itself.

I assume there is no version anywhere in the project depMgmt for `org.hibernate:hibernate-jpamodelgen` and so there is also no information which could map it to the new artifact `org.hibernate.orm:hibernate-jpamodelgen`. That mapping information is part of the old / original POM, but since the plugin does not have the version for this old artifact, it cannot lookup the POM with the relocation config.

Does that make sense? Or am I overlooking something in the setup?

Or this perhaps something that is working for you outside of the compiler plugin configuration with standard project dependencies?

**Comment by famod** [comment]
@psiroky 

Yeah, I think you have a very valid point: There is no explicit version for that old groupId and so there is no precise way to find the exact pom with the relocation info (=> new groupId).
As a human you could just check the latest version for the old groupId and get the relocation info from that but I doubt there is a nice way of finding the latest version via Maven builtin features (?).
Not worth the hassle.

For extra user-friendliness this plugin could check (if no version is found) for _similar_ deps, e.g. same artifactId but differen groupId and print put all those candidates, but relocations are not limited to groupIds (AFAIK), so...

It probably boils down to the question why there are/were no relocations warnings like you get for "normal" dependencies. For general plugin dependencies that's clearly something that Maven should cover.
`annotationProcessorPaths` is probably a different story.

**Comment by psiroky** [comment]
Yeah, there could possibly be some sort of heuristics to try to detect these relocations. Not sure if this would be something that should be done directly in the plugin or rather left to the dependency resolver.

I think the missing relocation warning has the same cause - the plugin (or more precisely the dependency resolution mechanism the plugin uses) simply does not know that the artifact was relocated, because without version, it cannot fetch the relocation config.

I believe the relocation warning gets printed in case you depend on the old artifact, but the version must be known, explicitly or via depMgmt: e.g:
```
<dependency>
  <groupId>org.hibernate</groupId>
  <artifactId>hibernate-jpamodelgen</artifactId>
  <verison>6.0.0</version>
```
In that case, the POM is fetched, relocation config found and the warning can be printed.

famod:
> I think the missing relocation warning has the same cause - the plugin (or more precisely the dependency resolution mechanism the plugin uses) simply does not know that the artifact was relocated, because without version, it cannot fetch the relocation config.
> 
> I believe the relocation warning gets printed in case you depend on the old artifact, but the version must be known, explicitly or via depMgmt: e.g:
> 
> ```
> <dependency>
>   <groupId>org.hibernate</groupId>
>   <artifactId>hibernate-jpamodelgen</artifactId>
>   <verison>6.0.0</version>
> ```
> 
> In that case, the POM is fetched, relocation config found and the warning can be printed.

I guess I should have been more precise:

I had exactly that (only a version property instead of the version literal, but that shouldn't make any difference) and still no warning. Maven 3.9.6, btw.

psiroky:
Ah, I see. In that case you are right, the relocation message should be there. I'll see if I can find some time to take a look at this.

jira-importer:
Resolve #627

PR #181, opened 2023-02-17
[MCOMPILER-381] - Refactor incremental detection

jorsol:

 - [X] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MCOMPILER) filed 
       pulling in other changes.
 - [X] Format the pull request title like `[MCOMPILER-XXX] - Fixes bug in ApproximateQuantiles`,
       commit message.

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
@@ -27,6 +27,8 @@ import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -59,11 +61,7 @@ import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.project.MavenProject;
 import org.apache.maven.shared.incremental.IncrementalBuildHelper;
 import org.apache.maven.shared.incremental.IncrementalBuildHelperRequest;
-import org.apache.maven.shared.utils.ReaderFactory;
 import org.apache.maven.shared.utils.StringUtils;
-import org.apache.maven.shared.utils.io.DirectoryScanResult;
-import org.apache.maven.shared.utils.io.DirectoryScanner;
-import org.apache.maven.shared.utils.io.FileUtils;
 import org.apache.maven.shared.utils.logging.MessageBuilder;
 import org.apache.maven.shared.utils.logging.MessageUtils;
 import org.apache.maven.toolchain.Toolchain;
@@ -83,6 +81,7 @@ import org.codehaus.plexus.compiler.util.scan.mapping.SourceMapping;
 import org.codehaus.plexus.compiler.util.scan.mapping.SuffixMapping;
 import org.codehaus.plexus.languages.java.jpms.JavaModuleDescriptor;
 import org.codehaus.plexus.languages.java.version.JavaVersion;
+import org.codehaus.plexus.util.FileUtils;
 import org.eclipse.aether.RepositorySystem;
 import org.eclipse.aether.artifact.Artifact;
 import org.eclipse.aether.artifact.ArtifactTypeRegistry;
@@ -560,12 +559,11 @@ public abstract class AbstractCompilerMojo extends AbstractMojo {
 
     /**
      * File extensions to check timestamp for incremental build.
-     * Default contains only <code>class</code> and <code>jar</code>.
      *
      * @since 3.1
      */
-    @Parameter
-    private List<String> fileExtensions;
+    @Parameter(defaultValue = "class,jar")
+    private Set<String> fileExtensions;
 
     /**
      * <p>to enable/disable incremental compilation feature.</p>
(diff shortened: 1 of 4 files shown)


slawekjaranowski:
@khmarbaise looks like you reported issue which this PR try to fix ... can you look?

slawekjaranowski (approved):

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -876,37 +874,33 @@ public void execute() throws MojoExecutionException, CompilationFailureException
...
+                Supplier<String> dependencyChanged = () -> isDependencyChanged() ? "changed dependency" : null;
+                Supplier<String> sourceChanged =
+                        () -> isSourceChanged(compilerConfiguration, compiler) ? "changed source code" : null;
+                Supplier<String> inputFileTreeChanged =
+                        () -> hasInputFileTreeChanged(computeInputFileTreeChanges(incrementalBuildHelper, sources))
+                                ? "added or removed source files"
+                                : null;
+
+                // Lazy evaluation of the incremental compilation detection.
+                String cause = Stream.of(immutableOutputFile, dependencyChanged, sourceChanged, inputFileTreeChanged)
```
slawekjaranowski, line 892: +1 for lazy evaluation

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -1156,6 +1151,16 @@ public void execute() throws MojoExecutionException, CompilationFailureException
 
             incrementalBuildHelper.beforeRebuildExecution(incrementalBuildHelperRequest);
 
+            // Cleanup the generated sources directory
+            if (getGeneratedSourcesDirectory() != null) {
```
olamy, line 1155: Why cleaning this directory? I don't understand the reason. Can you please explain?
reading the change as it it looks this will remove the generated-sources added via the phase `generate-sources` or any phase before compile. 
Maybe I read the change in a wrong way?

olamy (changes requested):

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -1400,10 +1406,11 @@ protected int getRequestThreadCount() {
         return session.getRequest().getDegreeOfConcurrency();
     }
 
-    protected Date getBuildStartTime() {
-        MavenExecutionRequest request = session.getRequest();
-        Date buildStartTime = request == null ? new Date() : request.getStartTime();
-        return buildStartTime == null ? new Date() : buildStartTime;
+    private Optional<Instant> getBuildStartTime() {
```
olamy, line 1409: uhm you're changing a protected method. this is breaking any backward compat.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -1539,36 +1546,41 @@ private static List<String> removeEmptyCompileSourceRoots(List<String> compileSo
      * generated classes and if we got a file which is &gt;= the build-started timestamp, then we caught a file which
      * got changed during this build.
      *
-     * @return <code>true</code> if at least one single dependency has changed.
+     * @return {@code true} if at least one single dependency has changed.
      */
-    protected boolean isDependencyChanged() {
-        if (session == null) {
+    private boolean isDependencyChanged() {
```
olamy, line 1551: breaking backward compat protected -> private

olamy (changes requested):

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
jorsol: I might be missing something, but I don't expect that anyone extend from AbstractCompilerMojo externally, I have never seen maven plugins as public APIs that can be extended, in other words protected works for subclasses that are on a different package, so this should not break anything.

The only consumer is m-compiler-p itself, and even if this method is used in a subclass it should be package-protected (no modifier) since the subclasses CompilerMojo and TestCompilerMojo are on the same package.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
jorsol: Same, this should have never been protected, is an internal method.

A good practice is to always start with the most restricted modifier (private) and then if needed open it to package-protected, then protected and finally public.

Again, this should not break anything since this class is not intended to be subclassed externally (unless I'm missing a weird use case and maven plugins can be used as public APIs).

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
olamy: try a github search and you will see https://github.com/search?q=extends+AbstractCompilerMojo&type=code
Even if it's maybe a bad idea this was not private from start but it is what it is we have to live (almost forever0 with that.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
olamy: same comment as here https://github.com/apache/maven-compiler-plugin/pull/181#discussion_r1404816887

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
jorsol: Actually there are many false-positive results, searching by  https://github.com/search?q=%22import+org.apache.maven.plugin.compiler.AbstractCompilerMojo%22&type=code gives a closer look of what is actually using this class, there are a few results, so yes, there are external consumers.

I will revert them to protected since it actually doesn't make a difference for what this PR is fixing.

BTW, forever is a long time, I expect this plugin and others to be rewritten someday and drop the legacy baggage ;-)

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
jorsol: Yes, let me explain, the generatedSourcesDirectory  is actually the directory for the annotation processor's generated source code:
https://github.com/apache/maven-compiler-plugin/blob/e5375fd10f00e4db5471bdc69ebdc4ba20f7bbdc/src/main/java/org/apache/maven/plugin/compiler/CompilerMojo.java#L114-L122

Since the "incremental" feature is actually a rebuild feature, this directory is generated by the javac compiler, it needs to be cleaned before the rebuild of the module or the build will fail with strange errors, this probably fixes many bugs related to the "incremental" compilation and annotation processors.

olamy: Oh right the method name or even field is kind of wrong but we can't change that

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -1577,31 +1589,35 @@ protected boolean isDependencyChanged() {
...
+            } catch (IOException ex) {
+                // we just cannot determine it, so don't do anything beside logging
+                getLog().warn("I/O error reading the lastModifiedTime: " + ex.getMessage());
             }
         }
 
         return false;
     }
 
+    private static String getFileExtension(Path path) {
```
olamy, line 1615: Don't we have already some plexus utils for that? Or the dependency have been removed?

olamy (changes requested):

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
jorsol: Plexus-utils is not included, there is one in maven-shared-utils but it's deprecated in favor of commons-io, yet commons-io is an indirect dependency, so we can make commons-io a direct dependency for that single method or simply duplicate the method here.

jorsol: I don't mind to make commons-io a direct dependency so will change this ASAP.

I'm not on a computer right now, so I will need some time to make all the requested changes.

Thanks for all the feedback! 👍

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
jorsol: Will need to undo the lazy evaluation, I plan to make more changes to the dependency detection and I need to store the status of the detection, if I lazily run this it will not execute and store the file and that means that the next execution could be recompiled again. 😞

slawekjaranowski (approved):

jorsol:
Hi @olamy, is there any other feedback?

jorsol:
Well, I think this should be ready to be merged if there is no other comment.

BTW, looking at the open issues, this should also fix [MCOMPILER-333](https://issues.apache.org/jira/browse/MCOMPILER-333) so it can be linked to this and declare it fixed for 3.12.0

jorsol:
> > Well, I think this should be ready to be merged if there is no other comment.
> > BTW, looking at the open issues, this should also fix [MCOMPILER-333](https://issues.apache.org/jira/browse/MCOMPILER-333) so it can be linked to this and declare it fixed for 3.12.0
> 
> it's usually better to have a PR per change, especially in case of non-related changes. But as you already hijacked this PR #186 here, it would be good to have this in the PR title and commits for history. And obviously, update https://issues.apache.org/jira/browse/MCOMPILER-333

Sure, will create a new PR for that change. the fix is simpler than the one from #186.

olamy:
`mvn spotless:apply` should help here ;)

jorsol:
> `mvn spotless:apply` should help here ;)

Done! 😅

PR #194, opened 2023-06-19
Remove references to old Maven versions.

elharo:
In 2023 assume everyone is on 3.3.1 or later.


slachiewicz:
I hope You remember discussion on ML about minimum supported Maven version in all our plugins - so far 3.2.5 is minimal for users https://maven.apache.org/docs/history.html#maven-3-6-x-and-before and You strongly voted against setting it to 3.5.4

elharo:
What we support and what we spend time calling out in documentation are two different things. Maven 2 is definitely dead.

slachiewicz:
`Remove references to old Maven versions. In 2023 assume everyone is on 3.3.1 or later.` if You say that this is true - so we can bump minimal supported Maven version for plugins to 3.3.1 and not be locked to 3.2.5 ;-)

michael-o (approved):
While I agree with @slachiewicz, I think we can do this and not waste more time for another PR in a year. People won't say thank you.

PR #195, opened 2023-06-19
[MCOMPILER-541] update maven-shared-utils to 3.4.2

elharo:


slawekjaranowski (approved):

slawekjaranowski:
Maybe add update or bump ... to commit message and jira to be more descriptive what we do

PR #200, opened 2023-10-04
[MCOMPILER-548] JDK 21 throws annotations processing warning that can not be turned off

hgschmie:
Turns out this is just a documentation issue. `full` works fine but is
not documented.

 - [X] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MCOMPILER-548) filed
       pulling in other changes.
 - [X] Format the pull request title like `[MCOMPILER-548] - Fixes bug in ApproximateQuantiles`,
       commit message.

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
@@ -294,7 +294,10 @@ public abstract class AbstractCompilerMojo extends AbstractMojo {
      * <ul>
      * <li><code>none</code> - no annotation processing is performed.</li>
      * <li><code>only</code> - only annotation processing is done, no compilation.</li>
+     * <li><code>full</code> - annotation processing and compilation.</li>
      * </ul>
+     *
+     * <code>full</code> is the default. Starting with JDK 21, this option must be set explicitly.
      *
      * @since 2.2
      * @see <a href="https://docs.oracle.com/en/java/javase/17/docs/specs/man/javac.html#option-proc">javac -proc</a>
--- a/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
@@ -303,7 +303,7 @@ public abstract class AbstractCompilerMojo extends AbstractMojo {
      * @see <a href="https://docs.oracle.com/en/java/javase/17/docs/specs/man/javac.html#option-proc">javac -proc</a>
      * @see <a href="https://docs.oracle.com/en/java/javase/17/docs/specs/man/javac.html#annotation-processing">javac Annotation Processing</a>
      */
-    @Parameter
+    @Parameter(property = "maven.compiler.proc")
     private String proc;
 
     /**


hgschmie:
Turns out that this might not be sufficient. JDK < 21 does *not* support "full" as a value, so we may want to check whether the value is "full" and omit it when the JDK is < 21. 

Requires more thought.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -279,8 +279,11 @@ public abstract class AbstractCompilerMojo extends AbstractMojo {
      * <ul>
      * <li><code>none</code> - no annotation processing is performed.</li>
      * <li><code>only</code> - only annotation processing is done, no compilation.</li>
+     * <li><code>full</code> - annotation processing and compilation.</li>
      * </ul>
      *
+     * <code>full</code> is the default. Starting with JDK 21, this option must be set explicitly.
```
bmarwell, line 285: What happens if I do not set it?  This seems to contradict the wording from line 276:

> If not set, both compilation and annotation processing are performed at the same time.

hgschmie: if you don't set it on JDK 21, you will be greeted with a warning that "soon, not setting proc:full will result in no longer doing annotation processing".

It is a very unsatisfying situation: 

JDK < 21: proc:full is not allowed, results in `error: invalid flag: -proc:full`
JDK == 21:  proc:full works, a warning is reported without it
JDK > 21: proc:full is *required* at some point, no more annotation processing without it

which puts the compiler plugin in the unfortunate situation of having to check the version of javac first before either not passing in `-proc:full` (because it is JDK < 21 and it is set by default) or passing it in for JDK >= 21.

That's why the comment reads "requires more thought". I have not found time to deal with it. Right now, I have a conditional profile in my poms... 😵‍💫

bmarwell: Another solution (which is probably intended by the JDK folks):
I used to do two executions in my projects:

* generate-sources: execute with `proc:only`
* default-compile: execute with `proc:none`

Worked like a charm and only changed it because Tamas hinted me to do so (because with this setup, some things like caching, incremental compilation etc. won't work that well or not at all).
However, this solution would not run into those problems at all.

cstamas: ... and I keep forgetting about this, and for me this sounds like "most correct" solution as well: split steps and use existing phases even we provide.

hgschmie: Interesting. Thank you for suggesting this. I can see the problems with incremental compilation but it still might be an acceptable workaround. Does that increase build times? Without static code analysis and tests, running the compiler is dominating build times and this seems to double those.

rmannibucau: +1 to keep backward compat and annot proc working by default.
Guess fastest is once the loader (urls) are know to cbexk if there is either the meta-inf SPI file or a module-info registering an annot proc and if so run peoc only before the compilation. Can be done on all versions probably.
Indeed, if explicitly disabled it is ignored.
So overall overhead is negligible and it will work for everyone and keep annlt proc goodness.

slawekjaranowski: we should add notice that `full` available for JDK 21+

EwoutH:
Any update on this effort? What's needed to move it forward?

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
rmannibucau: well issue is that `full` got backported to several lower versions too so I don't think we can write something relevant so ensuring we do the two passes when there is at least one SPI file sounds more robust until proc:none is set in the configuration (either `proc` or `compilerArgs`).

jddarcy: > well issue is that `full` got backported to several lower versions too so I don't think we can write something relevant so ensuring we do the two passes when there is at least one SPI file sounds more robust until proc:none is set in the configuration (either `proc` or `compilerArgs`).

In more detail, `-proc:full` will be supported in 11u and 17u (and Oracle 8u) as of the April 2024 releases: https://bugs.openjdk.org/browse/JDK-8308245 HTH

jddarcy: > Another solution (which is probably intended by the JDK folks): I used to do two executions in my projects:
> 
>     * generate-sources: execute with `proc:only`
> 
>     * default-compile: execute with `proc:none`
> 
> 
> Worked like a charm and only changed it because Tamas hinted me to do so (because with this setup, some things like caching, incremental compilation etc. won't work that well or not at all). However, this solution would not run into those problems at all.

Just running annotation processing when there is explicit need and intention to do so would be fine too.

The warning note is only printed when annotation processing by default is being used. If there is explicit configuration annotation processing related options, like `-processorpath` the note is not printed.

kwin: We can use conditions depending on the javac version (as we currently do in https://github.com/codehaus-plexus/plexus-compiler/blob/d8e98dda661b2e976f5382d43bb87dea1434381f/plexus-compilers/plexus-compiler-javac/src/main/java/org/codehaus/plexus/compiler/javac/JavacCompiler.java#L348). But IMHO m-compiler-p should add a more high level abstraction, as other compilers might handle that differently...

slawekjaranowski: we can add a property for `proc` parameter and property can be overridden in child modules with empty value.

rmannibucau: > We can use conditions depending on the javac version

fact is we don't know with javac version what is supported since several vendors did backport full support.

> we can add a property for proc parameter and property can be overridden in child modules with empty value.

It is already the case AFAIK (side note being adding properties for compiler args is quite misleading and lead to broken configuration more than it helps IMHO).

-> https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#proc

Think we don't have much choice to split the default compilation in 2 runs if there is at least one processor available (path or serviceloader discovery) and proc is not configured in compiler properties or args.

slawekjaranowski: this allow to override https://github.com/apache/maven-parent/pull/157

bmarwell: Then we can use a property for that.

My suggestion:
Leave it like in this PR for m-c-p 3. Detection is "smart" and will many builds, because they of course will have something like this already configured.

For m-c-p 4.x use two phases all the time, even if there's no APT. An empty run (no sources) doesn't hurt.

We can define properties for those as well, of course.

rmannibucau: Hmm, shouldnt we revert the parent change? except it is a breaking change it is also unexpected IMHO.

side note: this does not allow more than compilerArgs but enables to get conflicting args now ;).

Agree with you @bmarwell if v4 arrives at ~the same time than v3.x+1 since 21 starts to be used and everybody creates the verbose way in the pom which is very unlikely IMHO

slawekjaranowski (approved):

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
slawekjaranowski: I feel that we talk about one line in documentation too much 😄

PR #209, opened 2023-11-22
[MCOMPILER-558] compileSourceRoots in testCompile should be writable

lorenzsimon:
JIRA: https://issues.apache.org/jira/browse/MCOMPILER-558

Projects should be able to configure the `compileSourceRoots` also in the `testCompile` goal (like in the `compile` goal)

This causes a warning because the parameter is set as read-only.


slawekjaranowski:
@lorenzsimon - please preserve a line from PR template:

> I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)

in description

michael-o (approved):
Seems to be straight forward...

PR #222, opened 2024-02-03
(doc) Add new value accepted by `-proc:` option

pzygielo:
Without explicit `-proc`, `javac` complains about its future:

```
[INFO] Compiling 5 source files with javac [debug release 17] to target/test-classes
[INFO] Annotation processing is enabled because one or more processors were found
  on the class path. A future release of javac may disable annotation processing
  unless at least one processor is specified by name (-processor), or a search
  path is specified (--processor-path, --processor-module-path), or annotation
  processing is enabled explicitly (-proc:only, -proc:full).
  Use -Xlint:-options to suppress this message.
  Use -proc:none to disable annotation processing.
```

Current parameter's javadoc suggests that only `only` and `none` are allowed. This parameter however is not validated by m-compiler-p. This is also not validated by plexus-compiler.

-----


pzygielo:
Duplicate of
- #200

PR #231, opened 2024-03-12
[MCOMPILER-584] Refresh page - Using Non-Javac Compilers

slawekjaranowski:
https://issues.apache.org/jira/browse/MCOMPILER-584

### DIFF
--- a/src/site/apt/non-javac-compilers.apt.vm
+++ b/src/site/apt/non-javac-compilers.apt.vm
@@ -31,22 +31,24 @@ Using Non-Javac Compilers
 
   Contrary to this plugin's name, the Compiler Plugin does not compile the
   sources of your project by itself. To compile, the Compiler Plugin uses
-  another class to compile them. The parameter <<<compilerId>>> determines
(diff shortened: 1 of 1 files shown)


src/site/apt/non-javac-compilers.apt.vm
```diff
@@ -46,7 +48,7 @@ Using Non-Javac Compilers
 
   * <<<eclipse>>> with artifactId <<<plexus-compiler-eclipse>>>.
 
-  * <<<jikes>>> with artifactId <<<plexus-compiler-jikes>>>.
+  * <<<javac-with-errorprone>>>  with artifactId <<<plexus-compiler-javac-errorprone>>>.
```
CrazyHZM, line 51: javac-with-errorprone requires JDK 11+. Do you think this needs a special explanation?
https://codehaus-plexus.github.io/plexus-compiler/plexus-compilers/plexus-compiler-javac-errorprone/

slawekjaranowski: @CrazyHZM - requirements was added

CrazyHZM (approved):

PR #238, opened 2024-05-02
Exclude JDK 8 - temurin, adopt-openj9 on macos, use defaults values

slawekjaranowski:


elharo (approved):

.github/workflows/maven-verify.yml
```diff
@@ -28,7 +28,9 @@ jobs:
     with:
       # maven-args: "-Dinvoker.parallelThreads=2" cannot do this as this generate some concurrent download issues
       jdk-distribution-matrix: '[ "temurin", "zulu", "microsoft", "adopt-openj9" ]'
-      ff-jdk: "20"
-      jdk-matrix: '[ "8", "11", "17", "20" ]'
-      matrix-exclude: '[{ "jdk": "8", "distribution": "microsoft"},{ "jdk": "20", "distribution": "microsoft"},{ "jdk": "20", "distribution": "adopt-openj9"},]'
-      maven-matrix: '[ "3.6.3", "3.8.8", "3.9.5" ]'  # Maven versions matrix for verify builds
+      matrix-exclude: '[
```
olamy, line 31: the goal was to have a maximum of jdk to be tested as compiling can be highly dependant of the used jdk.
so with this we are really reducing the coverage.
could we keep testing on 11 maybe?

.github/workflows/maven-verify.yml
```diff
@@ -28,7 +28,9 @@ jobs:
...
+          { "jdk": "8", "distribution": "microsoft" },
```
elharo, line 32: Why is this one excluded? It's not a Mac

.github/workflows/maven-verify.yml
slawekjaranowski: done

.github/workflows/maven-verify.yml
slawekjaranowski: It has been excluded early ... probably microsoft distribution does not support jdk 8 at all

.github/workflows/maven-verify.yml
olamy: @slawekjaranowski Thanks! I was thinking of m-javadoc-p if we have similar "cleanup" :)

PR #294, opened 2025-02-09
Update docs for Maven 4 and release option

Bukama:
This PR updates several parts of the documentation to Maven 4 based on JDK 17 and aligns the docs to the new Java version schema. In particular on the front page it's mentioned that "8" is the default, where in the code it's actual "1.8" - but I wanted to keep docs consistent here.

Note: For Maven 3 branch I will create a separate PR later (maybe starting this evening)

### DIFF
--- a/src/site/apt/examples/set-compiler-source-and-target.apt.vm
+++ b/src/site/apt/examples/set-compiler-source-and-target.apt.vm
@@ -28,21 +28,25 @@
 
 Setting the <<<-source>>> and <<<-target>>> of the Java Compiler
 
+  (If you're are using version <<<3.13.0>>> or newer of the Compiler Plugin, use the recommended
+  {{{../examples/set-compiler-release.html}<<<maven.compiler.release>>> property}}.)
+
   Sometimes when you may need to compile a certain project to a different
   version than what you are currently using.  The <<<javac>>> can accept
   such command using <<<-source>>> and <<<-target>>>.  The Compiler Plugin
   can also be configured to provide these options during compilation.
+  You have to set the version following {{{https://openjdk.org/jeps/223}Java's new Version-String Scheme (JEP 223)}}}.
 
-  For example, if you want to use the Java 8 language features (<<<-source 1.8>>>)
-  and also want the compiled classes to be compatible with JVM 1.8 (<<<-target 1.8>>>),
-  you can either add the two following properties, which are the default property names for the plugin parameters:
+  For example, if you want to use the Java 8 language features and also want the compiled classes to be compatible
+  with JVM 8 (former 1.8) you can either add the two following properties, which are the default property names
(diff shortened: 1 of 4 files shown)


src/site/apt/index.apt.vm
```diff
@@ -29,8 +29,13 @@
 ${project.name}
 
   The Compiler Plugin is used to compile the sources of your project. The
-  default compiler used to compile Java sources is javax.tools.JavaCompiler.
-  Other compilers than <<<javac>>> can be used.
+  default compiler used to compile Java sources is <<<javac>>>.
+  If you want to use another compiler, refer to the {{{./non-javac-compilers} guide: Using Non-Javac Compilers}}.
```
slawekjaranowski, line 33: Can we use non javac compilers after: #271 @gnodet @desruisseaux ?

desruisseaux: Yes. Any compiler that implements the `javax.tool.JavaCompiler` interface should work. It is the case of Eclipse compiler for instance. I'm not sure if it has been tested however (not on my side), but we will fix issues or report them to ECJ if we found some.

src/site/apt/examples/set-compiler-source-and-target.apt.vm
```diff
@@ -28,21 +28,25 @@
 
 Setting the <<<-source>>> and <<<-target>>> of the Java Compiler
 
+  (If you're are using version <<<3.13.0>>> or newer of the Compiler Plugin, use the recommended
+  {{{../examples/set-compiler-release.html}<<<maven.compiler.release>>> property}}.)
+
```
desruisseaux, line 33: Note: this section may need to be updated later, after the work on `<sources>` element has been completed. The recommended approach would then be to specify the release together with the declaration of the sources.

No change needed now, it is just a reminder for revisiting later.

src/site/apt/index.apt.vm
slawekjaranowski, line 33: we need in link add `.html` to link `non-javac-compilers`

Bukama: Opps, thanks!

Bukama: > So should we update `non-javac-compilers` guide?

And shall I do it in this PR or do you (@desruisseaux) want to do it, when "the work on sources" (see other comment) is done?

slawekjaranowski: non-javac-compilers guide - in next PR, the best with IT

Bukama: > non-javac-compilers guide - in next PR, the best with IT

created https://issues.apache.org/jira/browse/MCOMPILER-609 for this.

slawekjaranowski (approved):

PR #311, opened 2025-03-11
Add the ggeneratedSourcesPath to the project source roots again

mensinda:
Followup for #191 to add back the `generatedSourcesPath` to the maven project paths.

This seems to work, but the solution feels a bit hacky.

---

Following this checklist to help us incorporate your
contribution quickly and easily:

- [x] Your pull request should address just one issue, without pulling in other changes.
- [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [x] Each commit in the pull request should have a meaningful subject line and body. 
  Note that commits might be squashed by a maintainer on merge.
- [x] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. 
  This may not always be possible but is a best-practice.
- [x] Run `mvn verify` to make sure basic checks pass.
  A more thorough check will be performed on your pull request automatically.
- [x] You have run the integration tests successfully (`mvn -Prun-its verify`).

If your pull request is about ~20 lines of code you don't need to sign an
[Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf) if you are unsure
please ask on the developers list.

To make clear that you license your contribution under
the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
you have to acknowledge this by using the following check-box.

- [x] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
- [ ] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

PR #191, opened 2023-06-15
[MCOMPILER-538] Do not add target/generated-sources/annotations to the source roots

mensinda:
Fix for [MCOMPILER-538](https://issues.apache.org/jira/browse/MCOMPILER-538).

The `generatedSourcesPath` will still be added to the *maven* project paths.

---

Following this checklist to help us incorporate your 
contribution quickly and easily:

 - [x] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MCOMPILER) filed 
       for the change (usually before you start working on it).  Trivial changes like typos do not 
       require a JIRA issue.  Your pull request should address just this issue, without 
       pulling in other changes.
 - [x] Each commit in the pull request should have a meaningful subject line and body.
 - [x] Format the pull request title like `[MCOMPILER-XXX] - Fixes bug in ApproximateQuantiles`,
       where you replace `MCOMPILER-XXX` with the appropriate JIRA issue. Best practice
       is to use the JIRA issue title in the pull request title and in the first line of the 
       commit message.
 - [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
 - [x] Run `mvn clean verify` to make sure basic checks pass. A more thorough check will 
       be performed on your pull request automatically.
 - [x] You have run the integration tests successfully (`mvn -Prun-its clean verify`).

If your pull request is about ~20 lines of code you don't need to sign an
[Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf) if you are unsure
please ask on the developers list.

To make clear that you license your contribution under 
the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
you have to acknowledge this by using the following check-box.

 - [x] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)

 - [x] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

PR #312, opened 2025-03-11
Add generatedSourcesPath back to the maven project

mensinda:
This is a followup for for #191 to add the generatedSourcesPath back to the maven project paths as the final step of the mojo.

This is an *alternative* implementation to #311.

---

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
@@ -703,8 +703,39 @@ public abstract class AbstractCompilerMojo extends AbstractMojo {
     private boolean targetOrReleaseSet;
 
     @Override
-    @SuppressWarnings("checkstyle:MethodLength")
     public void execute() throws MojoExecutionException, CompilationFailureException {
+        try {
+            executeReal();
+        } finally {
+            addGeneratedSourcesToProject();
+        }
+    }
+
+    private void addGeneratedSourcesToProject() {
+        File generatedSourcesDirectory = getGeneratedSourcesDirectory();
+        if (generatedSourcesDirectory == null) {
+            return;
+        }
+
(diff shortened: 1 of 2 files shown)


oehme:
LGTM

@slawekjaranowski This fixes the problem I mentioned

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -703,8 +703,39 @@ protected final Optional<Path> getModuleDeclaration(final Set<File> sourceFiles)
     private boolean targetOrReleaseSet;
 
     @Override
-    @SuppressWarnings("checkstyle:MethodLength")
     public void execute() throws MojoExecutionException, CompilationFailureException {
+        try {
+            executeReal();
+        } finally {
+            addGeneratedSourcesToProject();
+        }
```
slawekjaranowski, line 711: I'm not sure if finally is needed ... if something wrong in first method whole Maven build will be broken anymore

mensinda: Yes and no: I agree that the `try / finally` is not required for most circumstances (regular `mvn clean compile`, etc.). However, maven also has the `--fail-at-end` and `--fail-never` flags. In this case, the try finally could make sense.

So, just to be on the safe side to not change the original behavior, I would keep the `try / finally`. However, I am also open to removing it, if you would prefer it that way.

slawekjaranowski: ok, can stays

oehme: Try-finally is necessary, because there are several `return` points in the method and doing a finally block is the easiest way to ensure this work gets done, no matter how these `return`s are shuffled around in the future. It actually has nothing to do with failure.

slawekjaranowski (approved):

PR #273, opened 2024-12-19
[MCOMPILER-579] allow module-version configuration

mguillem:
Allow module-version configuration (like in master branch).

Following this checklist to help us incorporate your 
contribution quickly and easily:

 - [X] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MCOMPILER) filed 
       for the change (usually before you start working on it).  Trivial changes like typos do not 
       require a JIRA issue.  Your pull request should address just this issue, without 
       pulling in other changes.
 - [X] Each commit in the pull request should have a meaningful subject line and body.
 - [X] Format the pull request title like `[MCOMPILER-XXX] - Fixes bug in ApproximateQuantiles`,
       where you replace `MCOMPILER-XXX` with the appropriate JIRA issue. Best practice
       is to use the JIRA issue title in the pull request title and in the first line of the 
       commit message.
 - [X] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
 - [X] Run `mvn clean verify` to make sure basic checks pass. A more thorough check will 
       be performed on your pull request automatically.
 - [X] You have run the integration tests successfully (`mvn -Prun-its clean verify`).

If your pull request is about ~20 lines of code you don't need to sign an
[Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf) if you are unsure
please ask on the developers list.

To make clear that you license your contribution under 
the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
you have to acknowledge this by using the following check-box.

 - [X] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)

 - [ ] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

PR #331, opened 2025-05-26
Allow to not use --module-version for the Java compiler

pzygielo:

### DIFF
--- a/src/it/GH-331-no-module-version/invoker.properties
+++ b/src/it/GH-331-no-module-version/invoker.properties
@@ -1,0 +1,18 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+# 
+#   http://www.apache.org/licenses/LICENSE-2.0
+# 
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+invoker.java.version = 9+
--- a/src/it/GH-331-no-module-version/pom.xml
+++ b/src/it/GH-331-no-module-version/pom.xml
@@ -1,0 +1,45 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one
+  ~ or more contributor license agreements.  See the NOTICE file
+  ~ distributed with this work for additional information
+  ~ regarding copyright ownership.  The ASF licenses this file
+  ~ to you under the Apache License, Version 2.0 (the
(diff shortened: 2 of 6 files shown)


desruisseaux:
Do we need to add a new `useModuleVersion` plugin configuration option? Couldn't the users set the existing `moduleVersion` configuration option to an empty value instead if they do not want module version? It would still require a change in the code, but would avoid increasing the number of plugin configuration options.

pzygielo:
> Couldn't the users set the existing `moduleVersion` configuration option to an empty value instead if they do not want module version?

Let me explore that option. As long as the same goal is achieved I'm not attached to new parameter.

pzygielo:
> Do we need to add a new `useModuleVersion` plugin configuration option? Couldn't the users set the existing `moduleVersion` configuration option to an empty value 

Unclear how to preserve compatibility with current default of `project.version` (since 3.14). Using special value not conforming to `ModuleDescriptor.Version` could be considered, but I don't like that idea very much.

Thus I prefer proposed solution, but I'm open to see it being solved differently.

desruisseaux:
No special value, just allow users to specify `<moduleVersion/>`. The value is set to an empty or blank string, and the code just needs to test `if (!moduleVersion.isEmpty()) { add the option }`.

pzygielo:
With `<moduleVersion/>` `project.version` seems to be injected. So `!moduleVersion.isEmpty` would always be `true` unless `default` was removed.

slawekjaranowski (approved):

slawekjaranowski:
we can also use a special value like `skip` for `moduleVersion` parameter ... but I'm not sure if will be a clear way
so new parameter is ok for me

slawekjaranowski:
> I'm also not sure about special value.
> 
> But perhaps new option, if it's accepted, could have better name than I proposed here.

for me name is ok 😄

desruisseaux:
Thanks for the link. Alternatively, in the following declaration (added in commit 09dce4e613fbde5fbd88481b01d4a3004c31bea9):

```java
@Parameter(property = "maven.compiler.moduleVersion", defaultValue = "${project.version}")
private String moduleVersion;
```

What about removing the default value completely, and let users specify it explicitly? The rational given in [MCOMPILER-579](https://issues.apache.org/jira/browse/MCOMPILER-579) (that `${project.version}` is not a valid module version when the version ends with `-SNAPSHOT`) seems quite convincing.

slawekjaranowski:
> Thanks for the link. Alternatively, in the following declaration (added in commit [09dce4e](https://github.com/apache/maven-compiler-plugin/commit/09dce4e613fbde5fbd88481b01d4a3004c31bea9)):
> 
> ```java
> @Parameter(property = "maven.compiler.moduleVersion", defaultValue = "${project.version}")
> private String moduleVersion;
> ```
> 
> What about removing the default value completely, and let users specify it explicitly? The rational given in [MCOMPILER-579](https://issues.apache.org/jira/browse/MCOMPILER-579) (that `${project.version}` is not a valid module version when the version ends with `-SNAPSHOT`) seems quite convincing.

when we remove default value we will break a current behavior

pzygielo:
For the

> when we remove default value we will break a current behavior

additionally consider, that using the `project.version` was not introduced in
- #273.

It just enabled module version to be modified.
So the behaviour of _not specifying explicit module version and having it anyway_ is the old, old feature.

desruisseaux:
> (...snip...) is completely valid module version when it ends with anything,

Ah, okay. I do not remember having tested.

pzygielo:
Then, this could be merged as proposed, right?

Issue #334, opened 2025-06-03
DeltaList very inefficient for large projects


gsmet:
### Affected version

3.14.0

### Bug description

I'm experimenting with a large Quarkus project to try to enhance our build time and I noticed that `DeltaList` was very inefficient on this very large project.

It takes several seconds to figure out that nothing has changed when I compile the project again.

Here is a profile:

![Image](https://github.com/user-attachments/assets/02040be7-e03e-498d-a129-51c623b9c812)

As you can see, it represents close to 4% of the whole build (and the whole build is several dozens of seconds).

I'm experimenting with a patch.

desruisseaux:
The incremental compilation [is reported faster with Maven Compiler 4.x](https://github.com/apache/maven-compiler-plugin/issues/333#issuecomment-3257117699). Would it be possible to give it a try?

PR #335, opened 2025-06-03
Improve DeltaList behavior for large projects

gsmet:
This pull request targets the 3.x branch as I think this support has been rewritten in 4.

Here is the situation prior to this patch with a massive project (~40k source files):

Screenshot from 2025-06-03 17-17-35

> 7172 samples, 3.79%

The first commit keeps the ordering as is and make sure we write the file sorted and the outputs of `DeltaList` are sorted.
The behavior is exactly the same as before.

Here is the situation after the first commit:

Screenshot from 2025-06-03 17-17-27

> 78 samples, 0.04% - basically, the only thing left is the tiny thing you can see at the far right of the previous flame graph.

The second commit is a further optimization, which might not be useful and only sorts when displaying the output. The file is written unsorted.
It doesn't change much tbh so we might as well not include the second commit.
I did the work to be comprehensive so I thought I would present it.

Here is the situation after the second commit:

Screenshot from 2025-06-03 17-17-43

> 59 samples, 0.04%

**Opening as a draft so that we decide what to do about the second commit: my advice would be to drop it but YMMV**

BTW, it might be possible to optimize further but given this code is readable and give good results, I think it's good enough.

Fixes https://github.com/apache/maven-compiler-plugin/issues/334

----

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
@@ -1840,10 +1840,10 @@ public abstract class AbstractCompilerMojo extends AbstractMojo {
         }
         Path mojoConfigFile = mojoConfigBase.resolve(INPUT_FILES_LST_FILENAME);
 
-        List<String> oldInputFiles = Collections.emptyList();
+        Set<String> oldInputFiles = Collections.emptySet();
         if (Files.isRegularFile(mojoConfigFile)) {
             try {
-                oldInputFiles = Files.readAllLines(mojoConfigFile);
+                oldInputFiles = new HashSet<>(Files.readAllLines(mojoConfigFile));
             } catch (IOException e) {
                 // we cannot read the mojo config file, so don't do anything beside logging
                 getLog().warn("Error while reading old mojo status: " + mojoConfigFile);
@@ -1851,8 +1851,10 @@ public abstract class AbstractCompilerMojo extends AbstractMojo {
             }
         }
 
-        List<String> newInputFiles =
-                inputFiles.stream().sorted().map(File::getAbsolutePath).collect(Collectors.toList());
+        Set<String> newInputFiles = inputFiles.stream()
+                .sorted()
+                .map(File::getAbsolutePath)
+                .collect(Collectors.toCollection(LinkedHashSet::new));
 
         try {
             Files.write(mojoConfigFile, newInputFiles);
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/maven/plugin/compiler/DeltaList.java
```diff
@@ -33,10 +33,8 @@ final class DeltaList<E> {
     private final boolean hasChanged;
 
     DeltaList(Collection<E> oldList, Collection<E> newList) {
-        this.added = new ArrayList<>(newList);
-        this.removed = new ArrayList<>(oldList);
-        added.removeAll(oldList);
-        removed.removeAll(newList);
+        this.added = newList.stream().filter(i -> !oldList.contains(i)).collect(Collectors.toList());
+        this.removed = oldList.stream().filter(i -> !newList.contains(i)).collect(Collectors.toList());
```
gsmet: The idea is to use ``HashSet``s and avoid copying then removing a lot of elements.

jorsol: Could it be better to use the old for loop instead of streams?

gsmet: We could sure.

cstamas:
@desruisseaux @gnodet ping

gsmet:
I adjusted the commits:
- dropped the commit removing the sorts - this commit made things a bit faster but it's probably more future proof to keep the output properly sorted
- I switched from streams to for loops in `DeltaList`

cstamas (approved):

desruisseaux:
It looks fine to me as well.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -1840,19 +1840,21 @@ private boolean hasInputFileTreeChanged(IncrementalBuildHelper ibh, Set<File> in
...
                 return false;
             }
         }
 
-        List<String> newInputFiles =
-                inputFiles.stream().sorted().map(File::getAbsolutePath).collect(Collectors.toList());
+        Set<String> newInputFiles = inputFiles.stream()
+                .sorted()
+                .map(File::getAbsolutePath)
+                .collect(Collectors.toCollection(LinkedHashSet::new));
```
gnodet, line 1857: Would it be better to use a `TreeSet` instead of sorting + `LinkedHashSet` ?

gnodet: > Would it be better to use a `TreeSet` instead of sorting + `LinkedHashSet` ?

Quick lookup seems to indicate sorting + collecting the stream is slightly faster.

PR #937, opened 2025-06-21
Update the "non-javac-compilers" guide and add an integration test.

### LINKED ISSUE #904: [MCOMPILER-609] Update "non-javac-compilers"-guide
**[Matthias Bünger](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=JIRAUSER302100)** opened **[MCOMPILER-609](https://issues.apache.org/jira/browse/MCOMPILER-609?redirect=false)** and commented

From https://github.com/apache/maven-compiler-plugin/pull/294

The "non-javac-compilers"-guide should be updated (and an IT added)

> Yes. Any compiler that implements the javax.tool.JavaCompiler interface should work. It is the case of Eclipse compiler for instance. I'm not sure if it has been tested however (not on my side), but we will fix issues or report them to ECJ if we found some.

---
No further details from [MCOMPILER-609](https://issues.apache.org/jira/browse/MCOMPILER-609?redirect=false)

desruisseaux:
Fix #904.

### DIFF
--- a/src/it/MCOMPILER-609/src/main/java/myproject/HelloWorld.java
+++ b/src/it/MCOMPILER-609/src/main/java/myproject/HelloWorld.java
@@ -20,6 +20,6 @@ package myproject;
 
 public class HelloWorld {
     public static void main(final String[] args) {
-        System.out.println("Hi!");
+        System.out.println("Hello world");
     }
 }
--- a/src/site/apt/non-javac-compilers.apt.vm
+++ b/src/site/apt/non-javac-compilers.apt.vm
@@ -35,11 +35,11 @@ Using Non-Javac Compilers
 
   The parameter <<<compilerId>>> determines which class will be used.
 
-  By default, the Compiler Plugin uses the <<<javac>>> compiler
+  By default, the Maven Compiler Plugin uses the <<<javac>>> compiler
   bundled in the <<<javax.tools>>> package of the standard Java library.
-  But it is possible to use any other implementation, provided that the
+  But it is possible to use any other compiler, as long the
   implementation declares itself as a <<<javax.tools.JavaCompiler>>> service.
-  For using such implementation:
+  To use such an implementation:
 
   * Add the compiler implementation in the Maven compiler plugin dependencies.
 
@@ -49,7 +49,7 @@ Using Non-Javac Compilers
   []
 
   The example below shows how to use the Eclipse compiler <<<ecj>>>.
-  Note that at the contrary of Maven 3, there is no Plexus dependency to add.
+  Note: this configuration requires Maven 4 with Maven Compiler Plugin 4.
 
 +-----
 <project>


Bukama:
Hi, will have a look in the next days or next weekend. Modt prob during the next days, when at an unconference.

src/it/MCOMPILER-609/pom.xml
```diff
@@ -20,26 +20,29 @@ under the License.
...
     <plugins>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
-        <version>@pom.version@</version>
+        <version>@project.version@</version>
         <configuration>
-          <compilerId>eclipse</compilerId>
+          <compilerId>ecj</compilerId>
+          <!-- TODO: remove source and target after we identified where Maven inherits those values. -->
```
Bukama, line 36: Suggest to create an issue for this, so it's more visible

src/it/MCOMPILER-609/src/main/java/myproject/HelloWorld.java
```diff
@@ -16,10 +16,10 @@
...
  * under the License.
  */
-package org.example.pce.sample;
+package myproject;
 
-public class Dummy {
+public class HelloWorld {
     public static void main(final String[] args) {
-        String dummy = new Integer(1);
+        System.out.println("Hi!");
```
Bukama: `Hi!` is not `Hello world` :D

src/site/apt/non-javac-compilers.apt.vm
```diff
@@ -35,29 +35,21 @@ Using Non-Javac Compilers
 
   The parameter <<<compilerId>>> determines which class will be used.
 
-  Currently, the Compiler Plugin is bundled with the <<<javac>>> compiler
-  with artifactId <<<plexus-compiler-javac>>> version <<${plexusCompilerVersion}>>,
-  which can be seen as one of the dependencies declared inside the Compiler Plugin's POM.
+  By default, the Compiler Plugin uses the <<<javac>>> compiler
```
Bukama:
```suggestion
  By default, the Maven Compiler Plugin uses the <<<javac>>> compiler
```

src/site/apt/non-javac-compilers.apt.vm
```diff
@@ -35,29 +35,21 @@ Using Non-Javac Compilers
...
+  bundled in the <<<javax.tools>>> package of the standard Java library.
+  But it is possible to use any other implementation, provided that the
```
Bukama: Reads bumby, maybe this is a little less rocky?

> But it is possible to use any other compiler, as long the implementation declares itself as a ....

src/site/apt/non-javac-compilers.apt.vm
```diff
@@ -35,29 +35,21 @@ Using Non-Javac Compilers
...
+  implementation declares itself as a <<<javax.tools.JavaCompiler>>> service.
+  For using such implementation:
```
Bukama:
```suggestion
  To use such an implementation:
```

src/site/apt/non-javac-compilers.apt.vm
```diff
@@ -35,29 +35,21 @@ Using Non-Javac Compilers
...
   []
 
-* Using The Other Plexus Compilers
-
-  To use any of the non-javac compilers, you need to make the artifact available
-  in the classpath when the Compiler Plugin runs. This is possible by adding the
-  dependency when declaring the plugin in your project's <<<pom.xml>>>. The example
-  below shows how to use the <<<csharp>>> compiler:
+  The example below shows how to use the Eclipse compiler <<<ecj>>>.
+  Note that at the contrary of Maven 3, there is no Plexus dependency to add.
```
Bukama: Why does Plexus pops up here?

Bukama: Does this configuration is limited to Maven 3/4 or useable with both? Maybe should be noticed to be clear (esp. as some things are only limit to version 4.

Bukama (changes requested):
Overall good, only small suggestions/questions

src/site/apt/non-javac-compilers.apt.vm
desruisseaux: In Maven 3, it was necessary to add two dependencies:

* The Eclipse compiler
* A Plexus extension which was a wrapper around the Eclipse compiler.

Actually, the Plexus extension had the Eclipse compiler as a transitive dependency, so it may look as a single dependency in the `pom.xml`.

In Maven 4, there is no longer wrappers around the various compilers. The Eclipse compiler is used directly (through the standard `javax.tools.JavaCompiler` interface). I mentioned that in case some users were aware of the use of Plexus as an intermediate API, and were wondering whether Plexus was still needed. But I have no problem in omitting this details if too distracting.

desruisseaux: This configuration (without Plexus) works with the Maven Compiler Plugin 4 only, which itself requires Maven 4.

src/it/MCOMPILER-609/pom.xml
desruisseaux: Yes, the problem is that I have been unable to identify from where Maven inherits those values (`maven-core`? `maven-model-builder`? somewhere else?) NetBeans "effective" POM view does not show them.

src/site/apt/non-javac-compilers.apt.vm
desruisseaux: Added "Note: this configuration requires Maven 4 with Maven Compiler Plugin 4."

desruisseaux: Removed the mention of Plexus.

src/site/apt/non-javac-compilers.apt.vm
desruisseaux: Done, thanks for all the corrections.

Bukama (approved):
Thank you :)

PR #939, opened 2025-06-27
Add JDK 23 information and example for annotation processing

Bukama:
Since JDK 23 no annotation processing is done by default for security reason.
This PR adds information about this and also adds an example page how to do it in general and what is the recommended way to execute annotation processing (which applies also to earlier JDK versions).

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
@@ -18,15 +18,6 @@
  */
 package org.apache.maven.plugin.compiler;
 
-import javax.lang.model.SourceVersion;
-import javax.tools.DiagnosticListener;
-import javax.tools.JavaCompiler;
-import javax.tools.JavaFileManager;
-import javax.tools.JavaFileObject;
-import javax.tools.OptionChecker;
-import javax.tools.Tool;
-import javax.tools.ToolProvider;
-
 import java.io.BufferedReader;
 import java.io.BufferedWriter;
 import java.io.File;
@@ -51,7 +42,14 @@ import java.util.jar.Attributes;
 import java.util.jar.JarFile;
 import java.util.jar.Manifest;
 import java.util.stream.Stream;
-
+import javax.lang.model.SourceVersion;
+import javax.tools.DiagnosticListener;
+import javax.tools.JavaCompiler;
+import javax.tools.JavaFileManager;
+import javax.tools.JavaFileObject;
+import javax.tools.OptionChecker;
+import javax.tools.Tool;
+import javax.tools.ToolProvider;
 import org.apache.maven.api.JavaPathType;
 import org.apache.maven.api.Language;
 import org.apache.maven.api.PathScope;
@@ -78,7 +76,6 @@ import org.apache.maven.api.services.MessageBuilder;
 import org.apache.maven.api.services.MessageBuilderFactory;
 import org.apache.maven.api.services.ProjectManager;
 import org.apache.maven.api.services.ToolchainManager;
-
 import static org.apache.maven.plugin.compiler.SourceDirectory.CLASS_FILE_SUFFIX;
 import static org.apache.maven.plugin.compiler.SourceDirectory.MODULE_INFO;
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -51,7 +42,14 @@
 import java.util.jar.JarFile;
 import java.util.jar.Manifest;
 import java.util.stream.Stream;
-
+import javax.lang.model.SourceVersion;
```
Bukama: Imports where changed by `spotless:apply`

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -278,20 +275,26 @@ final Charset charset() {
...
+     *   <li>{@code none} – no annotation processing is performed, only compilation is done.</li>
      *   <li>{@code only} – only annotation processing is done, no compilation.</li>
-     *   <li>{@code full} – annotation processing and compilation are done.</li>
+     *   <li>{@code full} – annotation processing followed by compilation is done.</li>
      * </ul>
      *
-     * Prior Java 21, {@code full} was the default.
-     * Starting with Java 21, the default is {@code none} unless another processor option is used.
+     * The default value depends on the JDK used for the build.
+     * Prior to Java 22, the default was {@code full}, rso annotation processing and compilation were executed without explicit configuration.
```
desruisseaux: is 'rso" a typo? I guess the intend was "so"?

src/site/markdown/examples/annotationprocessor.md
```diff
@@ -0,0 +1,90 @@
...
+            <artifactId>maven-compiler-plugin</artifactId>
+            <version>...</version>
+            <configuration>
+                <annotationProcessorPaths>
+                    <path>
+                        <groupId>org.hibernate.orm</groupId>
+                        <artifactId>hibernate-processor</artifactId>
+                        <version>{version.hibernate}</version>
+                    </path>
+                </annotationProcessorPaths>
```
desruisseaux: This is the Maven 3 way. While it still work in Maven 4, it is marked as deprecated. The Maven 4 way would be to put the processor in `<dependencies>` with `<type>processor</type>`.

Should we copy this page in its current form to the 3.x branch, then edit this page on the `master` branch for the Maven 4 way?

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
Bukama: Yes it's a typo. Updated

Bukama:
@slawekjaranowski  Any idea why builds fails to spotless, even that I DO apply spotless?

```
[INFO] --- spotless:2.44.5:apply (default-cli) @ maven-compiler-plugin ---
[INFO] clean file: D:\Github\Maven\maven-compiler-plugin\src\main\java\org\apache\maven\plugin\compiler\AbstractCompilerMojo.java
[INFO] Spotless.Java is keeping 288 files clean - 1 were changed to be clean, 0 were already clean, 287 were skipped because caching determined they were already clean
[INFO] Spotless.Pom is keeping 128 files clean - 0 were changed to be clean, 0 were already clean, 128 were skipped because caching determined they were already clean
```

src/site/markdown/examples/annotation-processor.md
```diff
@@ -26,11 +26,41 @@ For example, the [Hibernate Processor](https://hibernate.org/orm/processor/) pro
...
+                        <artifactId>hibernate-processor</artifactId>
+                        <version>${version.hibernate}</version>
+                    </path>
+                </annotationProcessorPaths>
+            </configuration>
+        </plugin>
+        [...]
+    </plugins>
+</build>
+```
```
desruisseaux, line 60: The examples in other files seem to use an indentation of 2 spaces. Should we keep 2 spaces here for consistency?

Also, the other examples include the `<project>` enclosing element. Maybe we should keep it here as well, again for consistency.

src/site/markdown/examples/annotation-processor.md
```diff
@@ -26,11 +26,41 @@ For example, the [Hibernate Processor](https://hibernate.org/orm/processor/) pro
...
+
+### Maven 4
+Using Maven 4 and Maven Compiler Plugin 4.x, you can still use the same config as for Maven 3 and plugin version 3.x.
```
desruisseaux: Actually, the Maven 3 way is deprecated in the Maven 4 plugin. One reason for that is because the Maven 3 way does not provide control on class-path versus module-path. If we agree about this deprecation, and if we want to encourage users to migrate to the Maven 4 way, should the Maven 4 section be first?

desruisseaux: Another possible advantage of the Maven 4 way may be security. In the Maven 3 way, the compiler plugin is the only plugin to know which annotation processors were executed. In the Maven 4 way, all plugins can have this information. Therefore, a "bill of material" plugin could include this information in its report if desired.

Bukama: I updated the section, but left it below the Maven 3 section. I can totally not make a good assumption how fast Maven 4 will be adopted. I personally think quite quick, but many of the long time committer think it will take years.

src/site/markdown/examples/annotation-processor.md
Bukama: updated

desruisseaux (approved):

PR #942, opened 2025-07-02
Don't conflict source and release

fridrich:
Adding both --source and --release ends up in an error:
"error: option --source cannot be used together with --release"
So, ignore source and target specification if release is specified. The same behaviour as ant has.


desruisseaux:
Actually, it was also the behaviour of version 3.x of the compiler plugin. But it was intentionally removed in 4.x and documented in the [Incompatible changes](https://github.com/Geomatys/maven-compiler-plugin/wiki/Incompatible-changes) section (a pull request for porting that wiki to Maven documentation will be submitted later).

The rational is that version 3.x of the plugin applied a lot of heuristics: about `--source` versus `--release`, about class-path versus module-path, etc. The Oracle's developers decided that specifying both `--source` and `--release` is an error. Maybe their rational is that the user probably made a mistake, and not signalling that mistake may cause more problems later.  By adding an heuristic in Maven for hiding that error, we basically decide that Maven's own opinion prevails. I though that if we reduce as much as possible the situations where Maven's opinion diverges from OpenJDK opinion (in other words, if mapping to `javac` options is more straightforward in version 4.x of the plugin), we would hopefully reduce the risks of confusion by reducing "black magic" that users may not be aware of when comparing Maven's behaviour with e.g. tests executed directly with `javac` on the command line.

So in summary, the current behaviour is for better consistency with OpenJDK standard tools, and also because the standard `javac` behaviour is not necessarily bad: if there is a conflict between `--source` and `--release`, it may not be bad to require users to fix that inconsistency for avoiding unexpected results.

fridrich:
Just as a matter of fact, if you build and install the current maven-compiler-plugin and then modify it to use itself for building, you will hit that error. Why? Because it specifies the <release>${javaVersion}</release>, but there is source/target specified in the parent. For this particular case, removing that configuration fixes the stuff, but many times the problem is pretty far -- like in a parent of a parent of a parent -- and there the thing is complicated. OTOH, this is not my call, so whatever one decides must be fine with me.
Just having release parameter overpower source was quite consistent in the maven3 world and I don't see advantage of its removal other then educating maintainers. But it breaks stuff unnecessarily. But as I said, not my call.

desruisseaux:
Yes, there is an issue with `source` being inherited from parent `pom.xml` file somewhere. The current workaround, which is applied in the integration tests, is to declare a configuration like below:

```xml
<configuration>
  <!-- TODO: remove source and target after we identified where Maven inherits those values. -->
  <source />
  <target />
  <release>17</release>
</configuration>
```

I agree that this is ugly, but the real fix would be to fix the parent `pom.xml`. Since Maven 4 cannot be built with Java 8 anymore, there is no reason for keeping `--source` around. Having both `--source` and `--release` is a risk that someone updates `--source` without realizing that it will be ignored. Letting the standard `javac` tools raising the error as intended by OpenJDK developers may encourage us to do this cleaning.

PR #943, opened 2025-07-02
Port to maven 4.0.0-rc-4

fridrich:
Trivial changes so that the plugin builds against maven 4.0.0-rc-4

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/SourceDirectory.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/SourceDirectory.java
@@ -21,6 +21,7 @@ package org.apache.maven.plugin.compiler;
 import javax.lang.model.SourceVersion;
 import javax.tools.JavaFileObject;
 
+import java.nio.file.FileSystem;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.PathMatcher;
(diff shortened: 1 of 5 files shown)


src/main/java/org/apache/maven/plugin/compiler/SourceDirectory.java
```diff
@@ -287,8 +288,12 @@ static List<SourceDirectory> fromProject(
                 }
                 roots.add(new SourceDirectory(
                         directory,
-                        source.includes(),
-                        source.excludes(),
+                        source.includes().stream()
+                                .map(s -> FileSystems.getDefault().getPathMatcher(s))
```
desruisseaux: Instead of `FileSystems.getDefault()`, it should use `directory.getFileSystem()`. It is often the same file system, but not necessarily.

desruisseaux:
Also need to update the `.github/workflows/maven-verify.yml` file.

It is a small details, but:

```java
...
includes().stream().map(s -> directory.getFileSystem().getPathMatcher(s))
```

could be rewritten (for avoiding to invoke the same method many times, and also for shorter syntax):

```java
FileSystem fs = directory.getFileSystem();
...
includes().stream().map(fs::getPathMatcher)
```

PR #945, opened 2025-07-08
Make the default include/exclude syntax compatible with the behavior of Maven 3

desruisseaux:
This pull request temporarily copies the `PathSelector` class from Maven core with removal of unused features. A future version should use directly the `PathSelector` class from Maven core instead, but it may require that we move the latter in another ("shared"?) module.

The use of `PathSelector` allows the include/exclude patterns to reproduce the behavior of Maven 3 when no syntax is specified. Before this commit the default syntax was "glob", which is not identical to Maven 3 behavior.

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/PathSelector.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/PathSelector.java
@@ -1,0 +1,541 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.plugin.compiler;
+
+import java.io.File;
+import java.nio.file.FileSystem;
+import java.nio.file.Path;
+import java.nio.file.PathMatcher;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+
+/**
+ * Determines whether a path is selected according to include/exclude patterns.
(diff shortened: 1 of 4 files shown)


src/main/java/org/apache/maven/plugin/compiler/PathSelector.java
```diff
@@ -0,0 +1,541 @@
...
+    /**
+     * {@return a potentially simpler matcher equivalent to this matcher}.
+     */
+    @SuppressWarnings("checkstyle:MissingSwitchDefault")
+    public PathMatcher simplify() {
+        if (excludes.length == 0) {
+            switch (includes.length) {
+                case 0:
+                    return INCLUDES_ALL;
+                case 1:
```
fridrich, line 484: This case makes the thing like <includes><include>module-info.java</include></includes> not match the module-info.java in the compile source root. Once the case 1: is omitted, everything works correctly. Might be an unnecessary and wrong optimization, since the includes is not the only part of this. There are also directory includes, directory excludes and so on.

desruisseaux: Ah, good catch. I suspect that it is because when using the `includes[0]` matcher directly, `baseDirectory.relativize(path)` is not invoked.

fridrich: Yeah, and the optimization is not worth it besides being wrong. The case 0 is what we want, which now could be replaced by simple "if (includes.length == 0) return INCLUDES_ALL;"

fridrich: This would be something to change in maven 4 too, I guess. At least, here, all the separate module-info.java files were not compiled because of this. I mean, the situation when one compiles with release 8 the whole project and then compiles only the module-info.java in another execution. Those are typically cases when includes.length == 1. When I changed it to **/module-info.java, the logic made it two matchers and the simplify() was basically pass-through.

desruisseaux: It seems to me that the case 1 is the most common case, because the default matcher for compiling Java sources is `**/*.java`. The call to `baseDirectory.relativize(path)` is not necessary when the matcher starts with `**/`, so we could add that check in the case 1. It would cover all compilations with the default configuration. I do not know if the difference would be noticeable, but since it would cover millions of compilations around the world, it may be not that bad to have this small optimization.

fridrich: You are the one doing the PR. So, you get to decide. Just more complicated optimizations, more possibilities to get the thing wrong. If it was on me, I would go for code simplicity. But as I said, not my call :)

fridrich: But whatever decision, exactly the same thing should be done in maven, so that there is no subtle difference that will take again a day to understand :)

desruisseaux: Yes, it is my plan to port the change to Maven 4. There is already another bug fix (a `NullPointerException`) which needs to be ported anyway.

PR #950, opened 2025-07-16
Upgrade the dependencies of all remaining occurrences of commons-lang3.

desruisseaux:
Replace #949 in order to avoid to create separated pull requests for each occurrence of this dependency.


desruisseaux:
Note: #946 and #947 were about the same dependency.

PR #951, opened 2025-07-16
Two more test cases for multi-release integration tests

fridrich:
These two test cases helped us to detect errors while building multi-release jars with different module-info.java positions. Would be good to add them to the integration tests so that we don't regress.


desruisseaux:
Yes, more test cases are good. Launching the workflow.

src/it/multirelease-patterns/singleproject-root/pom.xml
```diff
@@ -0,0 +1,156 @@
...
+      </plugin>
+    </plugins>
+  </build>
+
+  <profiles>
+    <profile>
+      <id>jdk9</id>
+      <activation>
+        <jdk>[9,)</jdk>
+      </activation>
```
desruisseaux: Since we cannot compile Maven 4 with a release older than Java 17, this profile would always be activated. Do we need to keep it as a profile rather than directly in the plugin configuration?

src/it/multirelease-patterns/singleproject-separate-moduleinfo/pom.xml
```diff
@@ -0,0 +1,179 @@
...
+      </plugin>
+    </plugins>
+  </build>
+
+  <profiles>
+    <profile>
+      <id>jdk9</id>
+      <activation>
+        <jdk>[9,)</jdk>
+      </activation>
```
desruisseaux: Same remark as above about this profile being always activated.

src/it/multirelease-patterns/singleproject-root/pom.xml
desruisseaux: > Since we cannot compile Maven 4 with a release older than Java 17

I mean, with a Java older than 17 for running Maven. Not the value of the `<release>` parameter.

src/it/multirelease-patterns/singleproject-separate-moduleinfo/pom.xml
fridrich: working on it, just want to be sure all is fine before pushing

Issue #768, opened 2023-11-13
[MCOMPILER-556] debug=false still includes line numbers / documentation is misleading


jira-importer:
**[Marcono1234](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=marcono1234)** opened **[MCOMPILER-556](https://issues.apache.org/jira/browse/MCOMPILER-556?redirect=false)** and commented

## The issue

The [documentation of the `debug` parameter](https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#debug) says:

> Set to `true` to include debugging information in the compiled class files.

This suggests that setting it to `false` does not include debugging information. However, that is not actually the case. The class files still include debugging information, such as line numbers.

You can only get rid of debug information by setting `debug=true` and `debuglevel=none`, which is rather confusing.

As side note: Setting `debug=false` seems to have some effect because it does omit `LocalVariableTable` (and maybe more?) but it might still be confusing that it does not exclude all debug information.

(This might be a clone of MCOMPILER-114, but that is more than 10 years old and was closed back then during Jira clean-up, so it is not clear how much of that would still apply to the current state of the Maven Compiler Plugin.)

## Reproduction steps

(Tested with JDK 17 and Maven 3.9.1)

1. 

   Create a sample project with
   ----------------------------

   ```xml
        ```

   ```

\<project xmlns="http://maven.apache.org/POM/4.0.0"
xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
\<modelVersion>4.0.0\</modelVersion>

\<groupId>com.mycompany.app\</groupId>
\<artifactId>my-app\</artifactId>
\<version>1.0-SNAPSHOT\</version>

\<build>
\<pluginManagement>
\<plugins>
\<plugin>
\<groupId>org.apache.maven.plugins\</groupId>
\<artifactId>maven-compiler-plugin\</artifactId>
\<version>3.11.0\</version>
\<configuration>
\<release>11\</release>
\</configuration>
\</plugin>
\</plugins>
\</pluginManagement>
\</build>
\</project>
``` 
- 
```
class Test {
public void test() {
System.out.println("hello");
}
}
``` 
1. Run
``` 
mvn clean compile "-Dmaven.compiler.debug=false"
``` 
1. Inspect the compiled class file
``` 
[...]

PR #953, opened 2025-07-16
Deprecate the `debug` flag

desruisseaux:
The rational for this deprecation is discussed in #768.

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
@@ -440,16 +440,20 @@ public abstract class AbstractCompilerMojo implements Mojo {
     protected boolean debug = true;
 
     /**
-     * Keyword list to be appended to the {@code -g} command-line switch.
-     * Legal values are a comma-separated list of the following keywords:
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -432,7 +432,10 @@ final Charset charset() {
      *
      * @see #debuglevel
      * @see <a href="https://docs.oracle.com/en/java/javase/17/docs/specs/man/javac.html#option-g">javac -g</a>
+     *
+     * @deprecated Setting this flag to {@code false} is replaced by {@code <debuglevel>none</debuglevel>}.
```
slawekjaranowski, line 436: Should we also update description for `debuglevel` parameter?

desruisseaux: Done, thanks for the reminder.

desruisseaux:
Except for the camel-case remark, we received no more comment on this issue for almost 2 months. I presume that it is okay to merge the deprecation of `debuglevel`. It does not answer the question of whether we should remove that parameter completely, but the removal (if desired) can be a separated issue.

PR #958, opened 2025-09-10
Supoort command line paramenters and some optimizations

zhaoyunxing92:

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
@@ -546,7 +546,7 @@ public abstract class AbstractCompilerMojo implements Mojo {
     protected String outputTimestamp;
 
     /**
-     * The algorithm to use for selecting which files to compile.
+     * The algorithm to use for selecting which files to compile. The default values are implemented in `incrementalCompilationConfiguration()`
      * Values can be {@code dependencies}, {@code sources}, {@code classes}, {@code rebuild-on-change},
      * {@code rebuild-on-add}, {@code modules} or {@code none}.
      *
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -546,7 +546,7 @@ final Charset charset() {
     protected String outputTimestamp;
 
     /**
-     * The algorithm to use for selecting which files to compile.
+     * The algorithm to use for selecting which files to compile. The default values are implemented in `incrementalCompilationConfiguration()`
```
desruisseaux: The reader of this documentation cannot access easily the `incrementalCompilationConfiguration()` source code. I think that it is better to not add this reference.

desruisseaux:
It looks fine to me, with only a suggestion to remove the reference to `incrementalCompilationConfiguration()` package-private method in the documentation.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -546,7 +546,7 @@ final Charset charset() {
     protected String outputTimestamp;
 
     /**
-     * The algorithm to use for selecting which files to compile. The default values are implemented in `incrementalCompilationConfiguration()`
+     * The algorithm to use for selecting which files to compile,The default values are implemented in {@link #incrementalCompilationConfiguration()}
```
desruisseaux: The problem is that, in my understanding, this Javadoc will appear as the [plugin documentation](https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#useIncrementalCompilation). Readers will have no idea what this `incrementalCompilationConfiguration()` is. Even if we were generating classic Javadoc, this link would be broken because this field is protected and the `incrementalCompilationConfiguration()` method is package-private.

We have to remove this link from this javadoc. It was a tip for maintainers, not for users. If we want to keep this comment, it should be a comment on a line starting with `//` below the Javadoc.

PR #967, opened 2025-09-19
Follow Oracle Javadoc conventions

elharo:


desruisseaux:
If a pull request like this one (touching every source files) is created, I suggest to do it after #959 and #963 are merged (if accepted) for avoiding a large number of merge conflicts.

PR #975, opened 2025-10-02
Enable Java 25 and Maven 4 in CI

slachiewicz:


src/it/MCOMPILER-538/processor/pom.xml
```diff
@@ -36,6 +36,9 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <configuration>
           <proc>none</proc>
+          <annotationProcessors>
+            <annotationProcessor>foo.MyProcessor</annotationProcessor>
+          </annotationProcessors>
```
slawekjaranowski: to check if is needed

PR #979, opened 2025-10-06
Do not fail fast

cstamas:
By default the matrix will fail fast, but in this case it makes no sense, and is even better to make it non fail fast.

Ref:
https://github.com/apache/maven-gh-actions-shared/blob/3b6a3fd53a6cc1ea2f5f25625cc9fd6a9d6ebe22/.github/workflows/maven-verify.yml#L162


desruisseaux:
It looks fine to me. But wouldn't "fail fast" be nicer for the carbon footprint?

desruisseaux:
I'm closing this pull request on the basis of saving a little bit of carbon footprint and shared GHA runners. If the failure appears to be temporary, it is quite easy to ask GitHub to rerun only the jobs that did not completed.

PR #982, opened 2025-10-10
Fix java source and target version properties

gnodet:
Those properties are defined on the parent POM, inherited from maven-parent.
They are passed through to the forked maven on the command line by the failsafe plugin.
The current workaround is to remove those properties (and fix configuration for enforcer-plugin).

### DIFF
--- a/src/it/multirelease-patterns/multimodule/pom.xml
+++ b/src/it/multirelease-patterns/multimodule/pom.xml
@@ -38,7 +38,6 @@
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-compiler-plugin</artifactId>
           <version>@project.version@</version>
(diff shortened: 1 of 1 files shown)


src/it/multirelease-patterns/multimodule/pom.xml
```diff
@@ -38,11 +38,7 @@
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-compiler-plugin</artifactId>
           <version>@project.version@</version>
-          <configuration>
-            <!-- TODO: remove source and target after we identified where Maven inherits those values. -->
-            <source />
-            <target />
-          </configuration>
+          <configuration />
```
desruisseaux: Should we remove the configuration block completely?

PR #985, opened 2025-10-15
chore: drop junit3 references

sparsick:
In some cases, I delete the whole test class because the IT tests only test the compile goal.

### DIFF
--- a/src/it/MCOMPILER-275_separate-moduleinfo/pom.xml
+++ b/src/it/MCOMPILER-275_separate-moduleinfo/pom.xml
@@ -30,6 +30,14 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
   </properties>
 
+  <dependencies>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
+      <version>5.14.0</version>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
   <build>
     <plugins>
       <plugin>
--- a/src/it/MCOMPILER-275_separate-moduleinfo/src/test/java/com/foo/MyTest.java
+++ b/src/it/MCOMPILER-275_separate-moduleinfo/src/test/java/com/foo/MyTest.java
@@ -18,6 +18,15 @@
  */
 package com.foo;
 
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
 public class MyTest {
     private MyClass myClass = new MyClass();
+
+    @Test
+    public void testApp() {
+        assertTrue(true);
+    }
 }
--- a/src/it/MCOMPILER-284/src/test/java/com/foo/MyTest.java
+++ b/src/it/MCOMPILER-284/src/test/java/com/foo/MyTest.java
@@ -18,6 +18,15 @@
  */
 package com.foo;
 
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
 public class MyTest {
     private MyClass myClass = new MyClass();
+
+    @Test
+    public void testApp() {
+        assertTrue(true);
+    }
 }
--- a/src/it/default-incremental-disable/pom.xml
+++ b/src/it/default-incremental-disable/pom.xml
@@ -29,7 +29,14 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
   </properties>
-
+  <dependencies>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
+      <version>5.14.0</version>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
   <build>
     <plugins>
       <plugin>
--- a/src/it/includes-excludes/pom.xml
+++ b/src/it/includes-excludes/pom.xml
@@ -27,6 +27,14 @@ under the License.
   <description>Test that source includes/excludes with in the compiler plugin config.
     This will test excludes and testExcludes...</description>
 
+  <dependencies>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
+      <version>5.14.0</version>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
   <build>
     <plugins>
       <plugin>
--- a/src/it/includes-excludes/src/test/java/org/apache/maven/it0055/PersonTest.java
+++ b/src/it/includes-excludes/src/test/java/org/apache/maven/it0055/PersonTest.java
@@ -18,10 +18,18 @@
  */
 package org.apache.maven.it0055;
 
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
 public class PersonTest {
+
+    @Test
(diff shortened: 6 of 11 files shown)


src/it/MCOMPILER-275_separate-moduleinfo/pom.xml
```diff
@@ -44,7 +44,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
-        <version>@project.version@</version>
+        <version>3.14.1</version>
```
slawekjaranowski: Please revert, `@project.version@` is replaced by current version of plugin under test

sparsick: sorry for that. Fixed

src/it/default-fork-windows/src/test/java/MyTest.java
```diff
@@ -18,6 +18,4 @@
  */
 package foo;
 
-import junit.framework.TestCase;
```
desruisseaux, line 21: Actually, it would be nice to keep a dependency to something. It doesn't need to be JUnit 3. It can be JUnit 5 or anything else, even a dummy JAR. The goal is to test that a dependency declared with `<scope>test</scope>` is really on the classpath during the compilation of tests.

I would suggest to replace the JUnit 3 dependency by a JUnit 5 API dependency, and just put a dummy method with the `@Test` annotation. For example:

```java
import org.junit.jupiter.api.Test;

public class MyTest {
    @Test
    public void testSomething() {}
}
```

It should be sufficient for causing the test to fail if the compiler did not put the JUnit 5 dependency on the classpath, which is what we really want to test.

src/it/default-fork-windows/pom.xml
```diff
@@ -30,14 +30,6 @@
...
   </properties>
 
-  <dependencies>
-    <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
-      <version>3.8.2</version>
-    </dependency>
-  </dependencies>
-
```
desruisseaux, line 40: With above proposal to keep some dependency to a testing framework (any of them), it could be:

```xml
    <dependency>
      <groupId>org.junit.jupiter</groupId>
      <artifactId>junit-jupiter-api</artifactId>
      <version>5.14.0</version>
      <scope>test</scope>
    </dependency>
```

Or any other version of JUnit, or any other testing framework. The important thing is to have something with the `test` scope.

Note: I'm not sure why `<scope>test</scope>` was not present in the previous version of this test.

src/it/default-fork/src/test/java/MyTest.java
```diff
@@ -18,6 +18,4 @@
  */
 package foo;
 
-import junit.framework.TestCase;
```
desruisseaux, line 21: Same comment as above about keeping something (anything) available only under the `test` scope, for verifying that the compiler put the correct dependencies on the classpath.

src/it/default-fork-windows/src/test/java/MyTest.java
sparsick: Thanks for your feedback. I add Junit 5 deps to `default-fork` and `default-fork-windows`.

src/it/default-incremental-disable/pom.xml
```diff
@@ -30,14 +30,6 @@
...
   </properties>
 
-  <dependencies>
-    <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
-      <version>3.8.2</version>
-    </dependency>
-  </dependencies>
-
```
desruisseaux, line 40: Could we do that same here? I.e., replace JUnit 3 by JUnit 5, then add a dummy method annotated with `@Test` in the `MyTest` class below.

src/it/includes-excludes/src/test/java/org/apache/maven/it0055/PersonTest.java
```diff
@@ -18,14 +18,10 @@
  */
 package org.apache.maven.it0055;
 
-import junit.framework.TestCase;
-
```
desruisseaux, line 22: We could replace JUnit 3 dependency by JUnit 5, then replace above line by:

```java
import static org.junit.jupiter.api.Assertions.assertEquals;
```

which would allow us to keep the `assertEquals` in `PersonTest`. Here too, the real test is only to ensure that the compiler found the `assertEquals` method.

src/it/includes-excludes/src/test/java/org/apache/maven/it0055/PersonTest.java
```diff
@@ -18,14 +18,10 @@
...
-public class PersonTest extends TestCase {
+public class PersonTest {
     public void testPerson() {
         Person person = new Person();
 
         person.setName("foo");
-
-        assertEquals("foo", person.getName());
```
desruisseaux, line 29: We could keep those lines after the proposed addition of `import static org.junit.jupiter.api.Assertions.assertEquals`.

src/it/includes-excludes/src/test/java/org/apache/maven/it0055/PersonTwoTest.java
```diff
@@ -18,14 +18,10 @@
...
-
-public class PersonTwoTest extends TestCase {
+public class PersonTwoTest {
     public void testPerson() {
         Person person = new Person();
 
         person.setName("foo");
-
-        assertEquals("foo", person.getName());
     }
```
desruisseaux, line 34: Same as above: add `import static org.junit.jupiter.api.Assertions.assertEquals` and keep those lines as a way to ensure that JUnit was on the classpath.

src/it/includes-excludes/pom.xml
```diff
@@ -27,16 +27,6 @@ under the License.
...
-  <dependencies>
-    <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
-      <version>3.8.2</version>
-      <type>jar</type>
-      <scope>test</scope>
-    </dependency>
-  </dependencies>
-
```
desruisseaux, line 39: Replace by JUnit 5 dependency.

src/it/jdk16-annotation/src/test/java/com/mycompany/jdk16annotation/AppTest.java
```diff
@@ -18,35 +18,20 @@
...
-    public static Test suite() {
-        return new TestSuite(AppTest.class);
-    }
+    public AppTest(String testName) {}
 
     /**
      * Rigourous Test :-)
      */
-    public void testApp() {
-        assertTrue(true);
```
desruisseaux, line 50: Keep this line after adding a `import static org.junit.jupiter.api.Assertions.assertTrue`.

src/it/MCOMPILER-275_separate-moduleinfo/src/test/java/com/foo/MyTest.java
```diff
@@ -18,8 +18,6 @@
  */
 package com.foo;
 
-import junit.framework.TestCase;
```
desruisseaux, line 21: Replace by some dependency to JUnit 5. May be `@Test` or any `assert` method.

src/it/MCOMPILER-284/src/test/java/com/foo/MyTest.java
```diff
@@ -18,8 +18,6 @@
  */
 package com.foo;
 
-import junit.framework.TestCase;
```
desruisseaux, line 21: Replace by any dependency to JUnit 5.

PR #987, opened 2025-10-26
Annotation processor not always detected when using `<type>processor</type>`

desruisseaux:
In Maven 4, the recommended way to add a JAR on the annotation processor classpath is like below:

```xml
<project>
  <dependencies>
    <dependency>
      <groupId>org.apache.maven.plugins.compiler.it</groupId>
      <artifactId>annotation-processor</artifactId>
      <version>1.0-SNAPSHOT</version>
      <type>processor</type>
    </dependency>
  </dependencies>
</project>
```

However, Maven Compiler Plugin 4.0.0-beta-3 does not automatically detects the processor if it is declared only as above. This commit fixes that issue and adds an integration test, together with opportunistic cleaning in other integration tests.

In the main time, as a workaround, adding the following configuration seems to work, where `org.foo.MyAnnotationProcessor` is the value declared in the `META-INF/services/javax.annotation.processing.Processor` file. Note that only the processor class is declared, not the artifact coordinates:

```xml
<project>
  <build>
    <plugins>
      <plugin>
        <configuration>
          <annotationProcessors>
            <annotationProcessor>org.foo.MyAnnotationProcessor</annotationProcessor>
          </annotationProcessors>
        </configuration>
      </plugin>
    </plugins>
  </build>
</project>
```

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
@@ -1368,12 +1368,13 @@ public abstract class AbstractCompilerMojo implements Mojo {
         if (!executor.applyIncrementalBuild(this, configuration)) {
             return;
         }
-        Exception failureCause = null;
+        Throwable failureCause = null;
         final var compilerOutput = new StringWriter();
         boolean success;
         try {
             success = executor.compile(compiler, configuration, compilerOutput);
-        } catch (Exception e) {
+        } catch (Exception | NoClassDefFoundError e) {
+            // `NoClassDefFoundError` may happen if a dependency of an annotation processor is missing.
             success = false;
             failureCause = e;
         }
--- a/src/main/java/org/apache/maven/plugin/compiler/DiagnosticLogger.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/DiagnosticLogger.java
@@ -181,7 +181,7 @@ final class DiagnosticLogger implements DiagnosticListener<JavaFileObject> {
      *
      * @param cause if compilation failed with an exception, the cause
      */
-    Optional<String> firstError(Exception cause) {
+    Optional<String> firstError(Throwable cause) {
         return Optional.ofNullable(cause != null && firstError == null ? cause.getMessage() : firstError);
     }
 
--- a/src/main/java/org/apache/maven/plugin/compiler/ToolExecutor.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/ToolExecutor.java
@@ -654,6 +654,7 @@ public class ToolExecutor {
                 fileManager.setLocationFromPaths(StandardLocation.CLASS_OUTPUT, Set.of(outputForRelease));
                 latestOutputDirectory = outputForRelease;
                 unit.outputForRelease = outputForRelease;
+                sourcesForDebugFile.add(unit);
                 /*
                  * Compile the source files now. The following loop should be executed exactly once.
                  * It may be executed twice when compiling test classes overwriting the `module-info`,
@@ -681,7 +682,6 @@ public class ToolExecutor {
                     if (workaroundNeedsClose) {
                         workaround.close();
                     }
-                    sourcesForDebugFile.add(unit);
                     if (!success) {
                         break compile;
                     }


src/it/processor-type/annotation-user/pom.xml
```diff
@@ -0,0 +1,62 @@
...
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins.compiler.it</groupId>
+        <artifactId>annotation-verify-plugin</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+        <executions>
```
bmarwell, line 46: I have never seen this before...

src/it/CHANGES_v3_to_v4.md
```diff
@@ -0,0 +1,77 @@
...
+
+
+## Removed JUnit tests
+Removed the following directories and associated test methods:
+
+* `compiler-one-output-file-test2` because it was redundant with `compiler-one-output-file-test`.
+
+The only difference was the addition of include/exclude filters, but that difference had
+no effect because the compiler mock used in this test was ignoring all sources anyway.
+This test has been replaced by `compiler-modular-project`.
```
bmarwell, line 74: I think no one will care in the future about this. This is appropriate for this PR / issue to mention once, but for future use it is probably uninteresting.

src/it/CHANGES_v3_to_v4.md
```diff
@@ -0,0 +1,77 @@
...
+
+The tests in the following directores are not supported anymore and have been removed:
+
+* `release-without-profile` because the plugin no longer try to chose automatically
+   which parameters to use between `--source` and `--release`. This is justified by
+   the fact that the plugin cannot run on Java 8.
+* `release-without-profile-fork` for the same reason as above.
+* `MCOMPILER-190`, which has been replaced by `MCOMPILER-609`.
+  They are compilation tests using the Eclipse compiler, but the former test depended on Nexus.
+  It has been replaced by a test that depends on `javax.tools`.
```
bmarwell, line 64: I think no one will care in the future about this. This is appropriate for this PR / issue to mention once, but for future use it is probably uninteresting.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -1610,38 +1614,26 @@ final void resolveProcessorPathEntries(Map<PathType, List<Path>> addTo) throws M
...
-                if (annotationProcessorPaths == null || annotationProcessorPaths.isEmpty()) {
-                    DependencyResolver resolver = session.getService(DependencyResolver.class);
-                    if (resolver == null) { // Null value happen during tests, depending on the mock used.
-                        return false;
+            if (!strict || isVersionEqualOrNewer("RELEASE_23")) {
+                if (annotationProcessors == null || annotationProcessors.length == 0) {
+                    if (annotationProcessorPaths == null || annotationProcessorPaths.isEmpty()) {
+                        return dependencyTypes.contains(JavaPathType.PROCESSOR_CLASSES)
+                                || dependencyTypes.contains(JavaPathType.PROCESSOR_MODULES);
                     }
```
bmarwell: I know it has been there in a similar way before, but this is really hard to read. And instead of just adding comments, I (personally) would create at most one or two boolean methods where the method name alone explains what this if/if/if is all about.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -1559,11 +1568,6 @@ final DependencyResolverResult resolveDependencies(boolean hasModuleDeclaration)
      * Adds paths to the annotation processor dependencies. Paths are added to the list associated
      * to the {@link JavaPathType#PROCESSOR_CLASSES} entry of given map, which should be modifiable.
      *
-     * <h4>Implementation note</h4>
-     * We rely on the fact that {@link org.apache.maven.impl.DefaultDependencyResolverResult} creates
-     * modifiable instances of map and lists. This is a fragile assumption, but this method is deprecated anyway
-     * and may be removed in a future version.
-     *
```
bmarwell, line 1566: Well spotted!

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -1610,38 +1614,26 @@ final void resolveProcessorPathEntries(Map<PathType, List<Path>> addTo) throws M
      * {@return whether an annotation processor seems to be present}
      * This method is invoked if the user did not specified explicit incremental compilation options.
      *
+     * @param dependencyTypes the type of dependencies, for checking if any of them is a processor path
      * @param strict whether to be conservative if the current Java version is older than 23
```
bmarwell: Maybe not part of this PR, but we should really clarify what "strict" means and what conservative will change in the build.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
bmarwell: I was trying to think of an appropriate method here, but the first "if" needs extra explanation.

Maybe something along 

```java
if (isStrictJavaProcessingVersion() && containsAnnotationProcessorsInDependencies()) {}

// or

if (containsUsableAnnotationProcessorsInDependencies()) {} 
```

src/it/processor-type/annotation-user/pom.xml
desruisseaux: This is specific to the tests of the compiler plugin. This is a custom plugin defined in the setup phase. Actually, it is not really necessary because the same verification could be done in `verify.goovy`. I used it because other tests related to annotation processing already use it. If we replace by `verify.goovy`, we should do that for all.

src/it/CHANGES_v3_to_v4.md
desruisseaux: The purpose of this file is to document the differences between the 3.x and 4.0 branches. The `compiler-one-output-file-test2` directory is still present in the 3.x branch. Because some tests have been added in the 3.x branch without being ported yet to the 4.0 branch (e.g., [MCOMPILER-538](https://github.com/apache/maven-compiler-plugin/pull/191)), in the future we may want to compare the two branches for checking which tests present in one branch are missing in the other branch. This note is for remembering to not port that directory.

src/it/CHANGES_v3_to_v4.md
desruisseaux: As above note: this is for synchronization of integration tests between the two branches. I think those information could be deleted after the 3.x branch become unsupported.

src/it/processor-type/annotation-user/pom.xml
bmarwell: I don't mind, I was just curious!!

src/it/CHANGES_v3_to_v4.md
bmarwell: OK, I thought that should be obvious when browsing through the history, as the history would now have a reference to an issue. But now that you mention it, your commits do not have an issue linked, it seems.

desruisseaux: Actually, I'm lazily mixing two works in this same pull request. The work that really solves the issue described in this issue is the _"Fix: annotation processor not always detected when specified with <type>processor</type>"_ commit. But since this commit adds a new integration test, this pull request opportunistically does some cleanup in those tests first. Those cleanups are in separated commits, with the intend to not squash them. The `README.md` file is added as part of this cleanup, but it describes changes that were actually done long ago.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
desruisseaux: Just added a commit that tries to clarify. I added `isNullOrEmpty(…)` private methods, because they can be reused in about 10 places.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
desruisseaux: In the commit that I just pushed, I removed that flag. It was probably more conservative than necessary.

src/it/processor-type/annotation-processor/src/main/java/processor/SimpleAnnotationProcessor.java
```diff
@@ -0,0 +1,69 @@
...
+
+@SupportedSourceVersion(SourceVersion.RELEASE_17)
+@SupportedAnnotationTypes("user.SimpleAnnotation")
+public class SimpleAnnotationProcessor extends AbstractProcessor {
+    public SimpleAnnotationProcessor() {}
+
+    @Override
+    public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment roundEnv) {
+        // Verifies that transitive dependencies worked.
+        // TODO: pending a fix in Maven core.
```
Bukama: Please create an issue or this PR should be changed to draft-PR to don't miss this.

src/it/CHANGES_v3_to_v4.md
Bukama: I would rename the file to have the changes between v3 and v4 in name. That makes it more clear and pretends that other information are added to a "readme".

src/it/CHANGES_v3_to_v4.md
Bukama: > Actually, I'm lazily mixing two works in this same pull request. 

then why all those renames :D :D :D 💋

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -1451,13 +1458,37 @@ private boolean isVersionEqualOrNewer(String sourceVersion) {
         return supportedVersion.compareTo(requested) >= 0;
     }
 
+    /**
+     * Returns whether the given collection is null or empty, ignoring spaces.
+     * This is a convenience for a frequent check, and also for clarity.
+     */
+    private static boolean isNullOrEmpty(String c) {
```
Bukama: The method checks for `isBlank` and not `isEmpty` which is a difference

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -1608,42 +1634,25 @@ final void resolveProcessorPathEntries(Map<PathType, List<Path>> addTo) throws M
...
-                    DependencyResolverResult dependencies = resolver.resolve(DependencyResolverRequest.builder()
-                            .session(session)
-                            .project(project)
-                            .requestType(DependencyResolverRequest.RequestType.COLLECT)
-                            .pathScope(compileScope)
-                            .pathTypeFilter(allowedTypes)
-                            .build());
-
-                    return !dependencies.getDependencies().isEmpty();
+            if (isVersionEqualOrNewer("RELEASE_23")) {
```
Bukama: Can RELEASE_23 be replaced by a constant to avoid the literal here?

(I was hoping that there are some in the JDK, but according to https://docs.oracle.com/en/java/javase/17/docs/api/constant-values.html there are not)

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
Bukama: I'm with @bmarwell This is very hard to read :(

src/it/CHANGES_v3_to_v4.md
desruisseaux: It is because some integration tests were using the `install` goal. I replaced some `install` by `verify` [in this commit](https://github.com/apache/maven-compiler-plugin/commit/2f6b811981fd5badf5c0115e034362b6e68bc96c) but a few remain if I remember right. The use of `install` goal in integration tests pollutes the `~/.m2/repository/` directory. By using the `org.apache.maven.plugins.compiler.it` groupId for all tests, we ensure that if some `install` goals are used again for whatever reason, at least we have a single place to cleanup in the `.m2/repository` directory.

src/it/CHANGES_v3_to_v4.md
desruisseaux: Renamed as `CHANGES_v3_to_v4.md`.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
desruisseaux: This is updated in the latest commit.

src/it/processor-type/annotation-processor/src/main/java/processor/SimpleAnnotationProcessor.java
desruisseaux: It was my plan to create a pull request in Maven core next weekend. It will be a little bit of work, so waiting for the PR number would delay this PR by a week or so. But there is no rush, so I'm fine if we wait.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
desruisseaux: For the `groupId` renaming question, answer [here](https://github.com/apache/maven-compiler-plugin/pull/987#discussion_r2468334927).

For the "empty versus blank", I was taking "empty" is a lossy sense more similar to `Optional.isEmpty()`, for meaning "value is not specified", and I was hopping for a consistent name between the versions taking different kinds of arguments. In an attempt to clarify, I renamed all `isNullOrEmpty` by `isAbsent`.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
desruisseaux: Done. Also did the same for `RELEASE_22` for consistency.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
Bukama: Thanks. I think it‘s better to use such like isAbsent here to avoid the confusion 🙏

src/it/processor-type/annotation-processor/src/main/java/processor/SimpleAnnotationProcessor.java
desruisseaux: Done, replace by a link to https://github.com/apache/maven/pull/11373.

bmarwell (approved):

PR #988, opened 2025-11-02
JUnit Jupiter best practices

slachiewicz:
Use this link to re-run the recipe: https://app.moderne.io/recipes/org.openrewrite.java.testing.junit.JupiterBestPractices?organizationId=YTA5ODhiOTYtNDI5OS00OGY3LTg0NjctNGZiNmI4OTY1ZjY3

### DIFF
--- a/src/test/java/org/apache/maven/plugin/compiler/CompilerMojoTestCase.java
+++ b/src/test/java/org/apache/maven/plugin/compiler/CompilerMojoTestCase.java
@@ -54,6 +54,7 @@ import org.apache.maven.api.services.ToolchainManager;
 import org.apache.maven.impl.DefaultMessageBuilderFactory;
 import org.apache.maven.impl.InternalSession;
 import org.apache.maven.plugin.compiler.stubs.CompilerStub;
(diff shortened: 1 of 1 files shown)


slawekjaranowski:
and https://docs.junit.org/current/user-guide/#writing-tests-classes-and-methods

desruisseaux:
Actually their guide already raised my point:

> Another technical reason for making classes and methods public is to simplify testing on the module path when using the Java Module System.

Removing `public` is not really a problem, but an unnecessary complication in JPMS context. Therefore, the recommendation to omit `public` seems questionable to me.

olamy:
> Actually their guide already raised my point:
> 
> > Another technical reason for making classes and methods public is to simplify testing on the module path when using the Java Module System.
> 
> Removing `public` is not really a problem, but an unnecessary complication in JPMS context. Therefore, the recommendation to omit `public` seems questionable to me.

Agree.

slachiewicz:
Thanks for the feedback - hope that Maven will be soon modularized and also support plugin modules. 
Then we can run recipes to make it compatible. Would it also require some changes to the surefire plugin?

PR #991, opened 2025-11-10
Fix compatibility with Maven Core 4.1.0-SNAPSHOT

desruisseaux:
_**Edit:** a previous version of this pull request wrongly said that this issue exists since Maven core 4.0.0-rc-5. However after verification, it seems that https://github.com/apache/maven/pull/2505 has never been ported to the 4.0.x branch._

Since Maven core 4.1.0-SNAPSHOT, options formatted by `JavaPathType` have file names between quotes (https://github.com/apache/maven/pull/2505). This is desired for producing robust `target/javac.args` files allowing users to reproduce a compilation. However, the compiler plugin also has a `WorkaroundForPatchModule` class because (as of Java 24) `JavaFileManager` does not accept `StandardLocation.PATCH_MODULE_PATH` in calls to `setLocationForModule(…)`. That workaround expect file names without quotes. Therefore, the quotes added by Maven core needs to be removed.

Note: it does not necessarily means that Maven core should be modified, because `WorkaroundForPatchModule` is really (as its name implies) a workaround for a problem that should not exist.

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/ForkedToolSources.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/ForkedToolSources.java
@@ -470,7 +470,8 @@ final class ForkedToolSources implements StandardJavaFileManager {
      */
     void addAllLocations(List<String> command) {
         for (Map.Entry<PathType, Collection<? extends Path>> entry : locations.entrySet()) {
-            command.addAll(Arrays.asList(entry.getKey().option(entry.getValue())));
+            String[] options = entry.getKey().option(entry.getValue());
+            command.addAll(WorkaroundForPatchModule.optionWithoutQuotes(options));
         }
     }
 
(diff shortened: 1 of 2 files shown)


gnodet:
> _**Edit:** a previous version of this pull request wrongly said that this issue exists since Maven core 4.0.0-rc-5. However after verification, it seems that [apache/maven#2505](https://github.com/apache/maven/pull/2505) has never been ported to the 4.0.x branch._
> 
> Since Maven core 4.1.0-SNAPSHOT, options formatted by `JavaPathType` have file names between quotes ([apache/maven#2505](https://github.com/apache/maven/pull/2505)). This is desired for producing robust `target/javac.args` files allowing users to reproduce a compilation. However, the compiler plugin also has a `WorkaroundForPatchModule` class because (as of Java 24) `JavaFileManager` does not accept `StandardLocation.PATCH_MODULE_PATH` in calls to `setLocationForModule(…)`. That workaround expect file names without quotes. Therefore, the quotes added by Maven core needs to be removed.
> 
> Note: it does not necessarily means that Maven core should be modified, because `WorkaroundForPatchModule` is really (as its name implies) a workaround for a problem that should not exist.

Would it be better to backport #2505 ?
Or should the java compiler add the quotes maybe ?
Or maybe support both quoted and unquoted paths ? (and add /remove the quotes as needed) ?

Related to #11435

PR #997, opened 2025-11-16
Quote filenames written in shell scripts

desruisseaux:
Following https://github.com/apache/maven/pull/11435 (a change in Maven core for not quoting the files formatted by `JavaPathType` because filenames specified in `java.util.spi.ToolProvider` API shall *not* be quoted), quotes are added by the plugin instead, and only when the target is a shell script.

Opportunistically add a safety when path using only for debugging purposes cannot be relativized.

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
@@ -124,6 +124,12 @@ public abstract class AbstractCompilerMojo implements Mojo {
      */
     private static final String DEFAULT_EXECUTABLE = "javac";
 
+    /**
+     * The quote character for filenames in shell scripts.
+     * Shall not be used with {@link javax.tools.JavaFileManager}.
+     */
+    static final char QUOTE = '"';
+
     // ----------------------------------------------------------------------
     // Configurables
     // ----------------------------------------------------------------------
@@ -1775,7 +1781,11 @@ public abstract class AbstractCompilerMojo implements Mojo {
         final var commandLine = new StringBuilder("For trying to compile from the command-line, use:");
         Path dir = basedir;
         if (dir != null) { // Should never be null, but it has been observed with some Maven versions.
-            dir = Path.of(System.getProperty("user.dir")).relativize(dir);
+            try {
+                dir = Path.of(System.getProperty("user.dir")).relativize(dir);
+            } catch (IllegalArgumentException e) {
+                // Ignore, keep the absolute path.
+            }
             String chdir = dir.toString();
             if (!chdir.isEmpty()) {
                 boolean isWindows = (File.separatorChar == '\\');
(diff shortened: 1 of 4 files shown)


src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -1775,7 +1781,11 @@ private void writeDebugFile(final ToolExecutor executor, final Options configura
         final var commandLine = new StringBuilder("For trying to compile from the command-line, use:");
         Path dir = basedir;
         if (dir != null) { // Should never be null, but it has been observed with some Maven versions.
```
Bukama, line 1783: I don't get this comment in together with the if statement.
The comment says that dir should never be null, but it was observed, so I would assume the if handles the one path and an else path the rare, but observed, path if it is. But I don't see an else to handle this?

src/main/java/org/apache/maven/plugin/compiler/SourcePathType.java
```diff
@@ -102,7 +102,7 @@ public Optional<String> option() {
      */
     @Override
     public String[] option(Iterable<? extends Path> paths) {
-        var joiner = new StringJoiner(File.pathSeparator, (moduleName != null) ? moduleName + "=\"" : "\"", "\"");
+        var joiner = new StringJoiner(File.pathSeparator, (moduleName != null) ? moduleName + '=' : "", "");
```
Bukama, line 105: Here the quotes only got removed, but not replaced. Is this correct?

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
desruisseaux: `basedir` is a field declared in this class as below:

```java
@Parameter(defaultValue = "${project.basedir}", required = true, readonly = true)
protected Path basedir;
```

A value should always be injected by Maven core. But during the development of this plugin, sometime when upgrading from a Maven version to the next one, or when using a snapshot version compiled from `master`, the value was not supplied anymore by Maven core. For example, a similar issue can be reproduced with https://github.com/apache/maven/issues/11441 (I don't know if it was the same cause).

Even if a null `basedir` may be an indication of a serious problem that we want to be aware of, in this particular context, `basedir` is used only for making a path relative when creating a debug file. This code is executed only if the build failed (or was executed with `--verbose`), so we probably don't want the original cause to be hidden by a `NullPointerException` while writing debug information.

src/main/java/org/apache/maven/plugin/compiler/SourcePathType.java
desruisseaux: Yes, it needs to be consistent with https://github.com/apache/maven/pull/11435.

Bukama (approved):

PR #1022, opened 2026-01-27
Reorder cases to avoid skipping

elharo:


src/main/java/org/apache/maven/plugin/compiler/Options.java
```diff
@@ -320,13 +320,13 @@ private boolean checkNumberOfArguments(String option, int count, boolean immedia
         if (expected == count) {
             warning = null;
             return true;
-        } else if (expected < 1) {
+        } else if (expected == 0) {
```
elharo: This might still be wrong.

desruisseaux: We just need to replace `if (expected < 1)` by `if (expected < 0)`. This is the only correction needed (no need to change order or add another `if` condition).

elharo: The original code looks like this:

```
        } else if (expected < 1) {
            if (checker instanceof ForkedCompiler) {
                return true; // That implementation actually knows nothing about which options are supported.
            }
            warning = "The '" + option + "' option is not supported.";
        } else if (expected == 0) {
            warning = "The '" + option + "' option does not expect any argument.";
        } 
```

So if expected == 0 do we still need to check if checker instanceof ForkedCompiler?

desruisseaux: > So if expected == 0 do we still need to check if checker instanceof ForkedCompiler?

No, `ForkedCompiler` unconditionally returns -1 because it has no information about which options are supported by external tool.

elharo: done

elharo:
@retrigger

src/main/java/org/apache/maven/plugin/compiler/Options.java
desruisseaux: Sorry, I missed the comment saying that it was done. I Just had a look and it seems fine to me. Waiting for GitHub actions to complete and I will merge.

elharo:
CI is borked:   Error: No supported distribution was found for input adopt-openj9

elharo:
merged to master, let's see if that fixes the CI

desruisseaux:
It may be worth to check also if #1132 help.

PR #1027, opened 2026-02-02
Fixes #1028, try to make compiler summary message usable instead of obscure aggregates

### LINKED ISSUE #1028: DiagnosticListener summary message is obscur and misleading
### Affected version

4.0.0-beta-4

### Bug description

Current messages always logs:

- an aggregate
- code instead of pseudo english messages to be understandable
- is always logged whereas messages are already logged so if people care they have what they need, aggregate can be a fun way to encourage people to fix it but actually just having the log is sufficient in 99% of the case and it avoids the duplication

goal here is to

- make configurabe the output
- try to display more meaningful messages

rmannibucau:

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
@@ -1244,7 +1244,7 @@ public abstract class AbstractCompilerMojo implements Mojo {
      * @throws MavenException if an error occurred while fetching dependencies
      */
     public ToolExecutor createExecutor(DiagnosticListener<? super JavaFileObject> listener) throws IOException {
-        var executor = new ToolExecutor(this, useOrCreateListener(listener));
+        var executor = new ToolExecutor(this, listener);
         if (!(targetOrReleaseSet || executor.isReleaseSpecifiedForAll())) {
             MessageBuilder mb = messageBuilderFactory
                     .builder()
--- a/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
@@ -885,6 +885,16 @@ public abstract class AbstractCompilerMojo implements Mojo {
     @Parameter(property = "maven.compiler.maxmem")
     protected String maxmem;
 
+    /**
+     * Should maven log a summary of compiler messages (if any) by type.
+     * When there are a lot of messages of the same type (unchecked, deprecation etc),
+     * it can make the output more readable.
+     *
+     * @since 4.0.0-beta-5
+     */
+    @Parameter(property = "maven.compiler.messageLogType", defaultValue = "COMPILER")
+    protected MessageLogType messageLogType;
+
     // ----------------------------------------------------------------------
     // Read-only parameters
     // ----------------------------------------------------------------------
@@ -1227,14 +1237,14 @@ public abstract class AbstractCompilerMojo implements Mojo {
      * Therefore, the executor can safely be executed in a background thread,
      * provided that the {@link #logger} is thread-safe.
      *
-     * @param listener where to send compilation warnings, or {@code null} for the Maven logger
+     * @param listener where to send compilation messages, or {@code null} for the Maven logger
      * @return the task to execute for compiling the project using the configuration in this <abbr>MOJO</abbr>
      * @throws MojoException if this method identifies an invalid parameter in this <abbr>MOJO</abbr>
      * @throws IOException if an error occurred while creating the output directory or scanning the source directories
      * @throws MavenException if an error occurred while fetching dependencies
      */
     public ToolExecutor createExecutor(DiagnosticListener<? super JavaFileObject> listener) throws IOException {
-        var executor = new ToolExecutor(this, listener);
+        var executor = new ToolExecutor(this, useOrCreateListener(listener));
         if (!(targetOrReleaseSet || executor.isReleaseSpecifiedForAll())) {
             MessageBuilder mb = messageBuilderFactory
                     .builder()
@@ -1246,6 +1256,18 @@ public abstract class AbstractCompilerMojo implements Mojo {
             logger.warn(mb.build());
         }
         return executor;
+    }
+
+    /**
+     * Simple utility to enforce a listener.
+     * @param listener caller provided listener.
+     * @return user listener if not null else maven internal one.
+     */
+    private DiagnosticListener<? super JavaFileObject> useOrCreateListener(
+            final DiagnosticListener<? super JavaFileObject> listener) {
+        return listener == null
+                ? new DiagnosticLogger(logger, messageBuilderFactory, null, project.getRootDirectory(), messageLogType)
+                : listener;
     }
 
     /**
@@ -1370,7 +1392,7 @@ public abstract class AbstractCompilerMojo implements Mojo {
      */
     @SuppressWarnings("UseSpecificCatch")
     private void compile(final JavaCompiler compiler, final Options configuration) throws IOException {
-        final ToolExecutor executor = createExecutor(null);
+        final ToolExecutor executor = createExecutor(useOrCreateListener(null));
         if (!executor.applyIncrementalBuild(this, configuration)) {
             return;
         }
--- a/src/main/java/org/apache/maven/plugin/compiler/DiagnosticLogger.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/DiagnosticLogger.java
@@ -21,17 +21,21 @@ package org.apache.maven.plugin.compiler;
 import javax.tools.Diagnostic;
 import javax.tools.DiagnosticListener;
 import javax.tools.JavaFileObject;
+import javax.tools.ToolProvider;
 
 import java.nio.file.Path;
-import java.util.Arrays;
 import java.util.LinkedHashMap;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
+import java.util.ResourceBundle;
+import java.util.function.Consumer;
 
 import org.apache.maven.api.plugin.Log;
 import org.apache.maven.api.services.MessageBuilder;
 import org.apache.maven.api.services.MessageBuilderFactory;
+
+import static java.util.Optional.ofNullable;
 
 /**
  * A Java compiler diagnostic listener which send the messages to the Maven logger.
@@ -60,6 +64,11 @@ final class DiagnosticLogger implements DiagnosticListener<JavaFileObject> {
     private final Path directory;
 
     /**
+     * Type of output of the logger.
+     */
+    private final MessageLogType logType;
+
+    /**
      * Number of errors or warnings.
      */
     private int numErrors, numWarnings;
@@ -77,23 +86,30 @@ final class DiagnosticLogger implements DiagnosticListener<JavaFileObject> {
     /**
      * Creates a listener which will send the diagnostics to the given logger.
      *
-     * @param logger the logger where to send diagnostics
+     * @param logger                the logger where to send diagnostics
      * @param messageBuilderFactory the factory for creating message builders
-     * @param locale the locale for compiler message
-     * @param directory the base directory with which to relativize the paths to source files
+     * @param locale                the locale for compiler message
+     * @param directory             the base directory with which to relativize the paths to source files
+     * @param logType               output flavor
      */
-    DiagnosticLogger(Log logger, MessageBuilderFactory messageBuilderFactory, Locale locale, Path directory) {
+    DiagnosticLogger(
+            Log logger,
+            MessageBuilderFactory messageBuilderFactory,
+            Locale locale,
+            Path directory,
+            MessageLogType logType) {
         this.logger = logger;
         this.messageBuilderFactory = messageBuilderFactory;
         this.locale = locale;
         this.directory = directory;
+        this.logType = logType;
         codeCount = new LinkedHashMap<>();
     }
 
     /**
      * Makes the given file relative to the base directory.
      *
-     * @param  file  the path to make relative to the base directory
+     * @param file the path to make relative to the base directory
      * @return the given path, potentially relative to the base directory
      */
     private String relativize(String file) {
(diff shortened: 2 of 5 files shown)


src/main/java/org/apache/maven/plugin/compiler/ToolExecutor.java
```diff
@@ -72,11 +73,6 @@
  * @author Martin Desruisseaux
  */
 public class ToolExecutor {
-    /**
-     * The locale for diagnostics, or {@code null} for the platform default.
-     */
-    private static final Locale LOCALE = null;
-
```
desruisseaux, line 79: While this constant is null for now, its purpose is to make easier to find locale-sensitive codes in the future if we want to take locale in account. It may not happen in near term, but if it happens in the future, I would search for `Locale` or `LOCALE` on the whole code base.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -885,6 +885,16 @@ final void amendincrementalCompilation(EnumSet<IncrementalBuild.Aspect> aspects,
...
+    /**
+     * Should maven log a summary of compiler messages (if any) by type.
+     * When there are a lot of messages of the same type (unchecked, deprecation etc),
+     * it can make the output more readable.
+     *
+     * @since 4.0.0-beta-5
+     */
+    @Parameter(property = "maven.compiler.messageLogType", defaultValue = "COMPILER")
+    protected MessageLogType messageLogType;
+
```
desruisseaux, line 897: Could we use logger configuration instead of introducing a new plugin parameter? Currently, we use dependency injection like below, where `Log` is `org.apache.maven.api.plugin.Log`:

```java
@Inject
protected Log logger;
```

I think that each plugin should use a named logger. With https://github.com/apache/maven/pull/11685 recently merged, I would check if using `System.getLogger("org.apache.maven.compiler")` produces the desired output. If it works, then I suggest to drop `org.apache.maven.api.plugin.Log` and use system loggers or `java.util.logging` instead. Then:

* "org.apache.maven" would be the parent logger in Maven core and all plugins.
* "org.apache.maven.compiler" would be the logger for Maven Compiler Plugin.
* "org.apache.maven.compiler.summary" (or something else, proposal welcome) could be the logger for summary.

Logging can be configured in a logging framework specific way, but we could also add a `<logging>` section in the POM 4.1 model. We need to address Maven logging verbosity anyway, not just for the compiler plugin.

src/main/java/org/apache/maven/plugin/compiler/MessageLogType.java
```diff
@@ -0,0 +1,34 @@
...
+     */
+    ALL,
+    /**
+     * 1-1 between output and compiler messages.
+     */
+    COMPILER,
+    /**
+     * Only the summary.
+     */
+    SUMMARY,
```
desruisseaux, line 33: This mode duplicates the `-Xlint:none` compiler option. More generally, this enumeration duplicates logging configuration. See [above comment](https://github.com/apache/maven-compiler-plugin/pull/1027/changes#r2753718206) about an alternative that may be explored.

src/main/java/org/apache/maven/plugin/compiler/ToolExecutor.java
```diff
@@ -222,12 +218,8 @@ public class ToolExecutor {
     protected ToolExecutor(final AbstractCompilerMojo mojo, DiagnosticListener<? super JavaFileObject> listener)
             throws IOException {
 
+        this.listener = requireNonNull(listener, "DiagnosticListener can't be null in ToolExecutor");
```
desruisseaux, line 221: With this change, we are forcing all callers to create the `DiagnosticListener` themselves instead of creating the default instance in one single place. A `useOrCreateListener` method was added in another change, but it forgot that `ToolExecutor` is extended by `ToolExecutorForTest`. The latter is now broken (**EDIT:** with a second though, it is not broken because of a duplicated call to `useOrCreateListener` in the private `compiler` method).

If the goal was to access the `messageLogType` field, it was not necessary to move this code. It can be accessed as `mojo.messageLogType`.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -1370,7 +1392,7 @@ public Options parseParameters(final OptionChecker compiler) {
      */
     @SuppressWarnings("UseSpecificCatch")
     private void compile(final JavaCompiler compiler, final Options configuration) throws IOException {
-        final ToolExecutor executor = createExecutor(null);
+        final ToolExecutor executor = createExecutor(useOrCreateListener(null));
```
desruisseaux, line 1395: This call to `useOrCreateListener` duplicates the same call done in `createExecutor`.

**EDIT:** with a second though, it is a duplication for the main code, but not for `TestCompilerMojo`.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -1248,6 +1258,18 @@ public ToolExecutor createExecutor(DiagnosticListener<? super JavaFileObject> li
...
     }
 
+    /**
+     * Simple utility to enforce a listener.
+     * @param listener caller provided listener.
+     * @return user listener if not null else maven internal one.
+     */
+    private DiagnosticListener<? super JavaFileObject> useOrCreateListener(
+            final DiagnosticListener<? super JavaFileObject> listener) {
+        return listener == null
```
desruisseaux, line 1268: I prefer that we keep this construction inside `ToolExecutor` as today, in order to have a single place where the default listener is constructed. It would avoid my confusion in [this](https://github.com/apache/maven-compiler-plugin/pull/1027/changes#r2753790894) and [this](https://github.com/apache/maven-compiler-plugin/pull/1027/changes#r2753796792) comment.

src/main/java/org/apache/maven/plugin/compiler/ToolExecutor.java
rmannibucau: I understand but as of today it is more misleading since it looks like it is supported and buggy cause injected statically whereas it is dead code so I preferred to drop it

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
rmannibucau: I thought about that and rational was that this is a behavior you want to have and not logger related, this is closer to an "output format" than a channel with a log level. Also think it would be a bad practise to abuse logger as we did abuse type in terms of design cause loggers are intended for output management and not for output formatting/content selection primarily - even if we can always workaround things.

so think we are good this way

src/main/java/org/apache/maven/plugin/compiler/MessageLogType.java
rmannibucau: I dont understand that at all:

* logging -> see previous answer
* duplication -> it is already the case, if you dislike it I'm happy to drop summary feature at all too, it fixes current issues

src/main/java/org/apache/maven/plugin/compiler/ToolExecutor.java
rmannibucau: yes but I think it is cleaner to ensure we have a listener and avoid to rely on mojo there
the issue having the default there is that it makes the code particular to a single use case
I hesitate to drop tool executor to bring it back in the abstract mojo to keep the coupling

long story short, I think it doesn't change much thing even if we should probably clean up the tool executor coupling and keep it more generic (as plexus was before)

src/main/java/org/apache/maven/plugin/compiler/DiagnosticLogger.java
```diff
@@ -191,50 +211,93 @@ Optional<String> firstError(Throwable cause) {
...
-    /**
-     * Appends the count of warnings or errors, making them plural if needed.
-     */
-    private static void writeCount(MessageBuilder message, String patternForCount, int count, String name) {
-        message.newline();
-        message.format(patternForCount, count, name);
-        if (count > 1) {
-            message.append('s');
+    // we mainly know the one for javac as of today, this impl is best effort
+    private ResourceBundle tryGetCompilerBundle() {
```
desruisseaux, line 273: This is very specific to the implementation details of the OpenJDK compiler, and I don't think that it will work in JPMS context (access to non-exported resources are blocked by the JVM). It is not exactly the desired properties file anyway, because the values are for example _"{0} uses or overrides a deprecated API"_ and we have nothing to provide for the `{0}` part. We may have to provide our own `.properties` file. It would cover only a subset of the messages, but we could cover the most common ones.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
rmannibucau: just tried to deduplicate the useOrCreateListener call

src/main/java/org/apache/maven/plugin/compiler/DiagnosticLogger.java
rmannibucau: > I don't think that it will work in JPMS context

It does in our context, give it a try

> It is not exactly the desired properties file anyway

Not sure what you mean cause it is literally the pattern message for the code you used :thinking: , so the intent and since it is an aggregation we don't want to fill the placeholders there, but at least the user sees something readable

src/main/java/org/apache/maven/plugin/compiler/ToolExecutor.java
desruisseaux: It is neither buggy nor misleading. It is supported: if someone puts a value different then null, it will be taken in account. The fact that it is `static final` clearly indicates that it is not configurable yet. It is not dead code, it is a constant in the same was as if we defined a constant of 0. Please revert.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
desruisseaux: This is not output formatting, but output filtering, which is in scope of logger's configuration.

src/main/java/org/apache/maven/plugin/compiler/MessageLogType.java
desruisseaux: With this pull request, in `DiagnosticListener.report` if the mode is `SUMMARY`, the compiler error or warning is not logged. This is equivalent to running `javac` with `-Xlint:none` option, which disable all warnings.

One difference between `SUMMARY` and `-Xlint:none` is that `SUMMARY` disables not only the warnings but also the errors, which is probably not desired.

src/main/java/org/apache/maven/plugin/compiler/DiagnosticLogger.java
desruisseaux: > It does in our context, give it a try

Yes but this is not a JPMS context. While Maven is not yet modularized, this is one of my long term goal.

> Not sure what you mean cause it is literally the pattern message for the code you used

Yes but with parameters that we don't have when writing a summary. The "_{0} uses or overrides a deprecated API"_ is a message reported when one particular class or method use a deprecated API, and "{0}" is the name of that class or method. But when we write the summary, we are saying that (for example) this warnings occurred 100 times. We have no particular class or method that we can insert in the "{0}" part.

src/main/java/org/apache/maven/plugin/compiler/ToolExecutor.java
rmannibucau: I didnt say it was buggy but it looks like that - was my first thought checking the code.

> if someone puts a value different then null, it will be taken in account

then it will be a bug so we must not do it, it can be acceptable to make it not static but it would stay dead code cause it is never written - doesn't idea warns about it btw, it should?

So think it is really safer and saner to not do it until we do it cleanly cause it will also require to handle locale configuration and think about doing it globally or per execution.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
rmannibucau: filtering for a logger is only the level which doesn't apply well at all there

agree "formatting" is poor from a semantic point of view but still, it doesn't match logger design

also think we should stick to logger=class until we abuse MDC globally which is not needed there and would require a lot of rework

src/main/java/org/apache/maven/plugin/compiler/MessageLogType.java
rmannibucau: it disables the details yes and if you have an error you rerun with `-D` to see it, think it matches more the summary use case than the opposite

I'm also not sure you're right saying it is equivalent to `-Xlint:none` since summary will be empty with that as well, no? :thinking:

src/main/java/org/apache/maven/plugin/compiler/DiagnosticLogger.java
rmannibucau: > While Maven is not yet modularized, this is one of my long term goal.

hmm, there is no consensus on that so tempted to say it will not happen so it is ok, worse case it would be easy to make it working too so not a blocker IMHO

src/main/java/org/apache/maven/plugin/compiler/ToolExecutor.java
desruisseaux: > then it will be a bug so we must not do it, it can be acceptable to make it not static but it would stay dead code

If someone puts `Locale.JAPANESE`, some texts will appear in Japanese, which is the expected behaviour, not a bug. It is not dead code neither, it is the declaration of a constant (for now), and `null` is a perfectly valid constant value.

Configuration may come in the future, which is precisely the reason why this constant is defined: if some day we want to support `Locale` configuration, we will just need to remove the `static final` part. The purpose of this constant is to save time in the future by avoiding the need to search where a locale is used. With the change in this pull request, `LOCALE` is replaced by `null` in method call arguments. It is not easy to guess by reading the code that the argument was a locale. With the `LOCALE` constant, it is obvious.

Please revert. It is not because the value of a constant is zero or null that it is dead code.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
desruisseaux: > filtering for a logger is only the level

In `java.util.logging`, filtering is also by logger name. Filtering is also by handler (different filtering for logs sent to the console or logs sent to a file). I never use SLF4J, but I presume that it has the same capability than `java.util.logging`. If we use system logging, it will delegate either to `java.util.logging`  (default), or to SLF4J, therefore is would have this capability as well.

> also think we should stick to logger=class

`java.util.logging` uses a more generic term, "subsystem", with a recommendation to base it on the package name or class name. In other words, the logger name should indeed starts at least with the package name, but there is some flexibility for what come after the package name.

src/main/java/org/apache/maven/plugin/compiler/ToolExecutor.java
desruisseaux: > yes but I think it is cleaner to ensure we have a listener

`DiagnosticListener` is used only by `javax.tools.JavaCompiler` in tasks that are created only by `ToolExecutor`. The logic was that the `ToolExecutor` constructor is a mandatory passage before the listener is used, therefore creating a default listener in `ToolExecutor` constructor guarantees that we have a listener in the class that needs it. By contrast, `AbstractCompilerMojo` does nothing with the listener.

src/main/java/org/apache/maven/plugin/compiler/DiagnosticLogger.java
desruisseaux: > hmm, there is no consensus on that so tempted to say it will not happen so it is ok, worse case it would be easy to make it working too

Agree that modularizing Maven is not on the agenda soon. However, it will not be easy at all to make the pull request works in that case. It will be blocked by the JVM, unless we put `--add-opens` options in the Maven launch script, which is strongly discouraged except for testing. Generally, using internal API is discouraged even in a non-JPMS world, and it is precisely for avoiding such practices that JPMS makes it hard.

Furthermore, we still have the issue of resource strings not applicable (the problem of having no value to supply in "{0}" emplacements).

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -885,6 +885,16 @@ final void amendincrementalCompilation(EnumSet<IncrementalBuild.Aspect> aspects,
     @Parameter(property = "maven.compiler.maxmem")
     protected String maxmem;
 
+    /**
+     * Should maven log a summary of compiler messages (if any) by type.
```
elharo, line 889: Summarize compiler messages by type.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -885,6 +885,16 @@ final void amendincrementalCompilation(EnumSet<IncrementalBuild.Aspect> aspects,
...
+     * When there are a lot of messages of the same type (unchecked, deprecation etc),
+     * it can make the output more readable.
```
elharo, line 891: it can make --> this makes

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
elharo: I tend to think this belongs in Maven, not the logger

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -1248,6 +1258,18 @@ public ToolExecutor createExecutor(DiagnosticListener<? super JavaFileObject> li
         return executor;
     }
 
+    /**
+     * Simple utility to enforce a listener.
+     * @param listener caller provided listener.
```
elharo, line 1263: blank line

src/main/java/org/apache/maven/plugin/compiler/DiagnosticLogger.java
```diff
@@ -118,63 +134,67 @@ public void report(Diagnostic<? extends JavaFileObject> diagnostic) {
...
+            MessageBuilder record = messageBuilderFactory.builder();
+            record.a(message);
+            JavaFileObject source = diagnostic.getSource();
+            Diagnostic.Kind kind = diagnostic.getKind();
+            String style;
+            switch (kind) {
+                case ERROR:
+                    style = ".error:-bold,f:red";
+                    break;
+                case MANDATORY_WARNING:
```
elharo, line 147: fallthrough is error prone

elharo (changes requested):

slachiewicz:
The last CI run on this branch was 2026-02-02, and GitHub has deleted the logs since, so there's no way to tell whether the change still builds against master.

A rebase triggers a fresh run and makes the state readable again. If you'd rather not carry it forward, closing is equally useful — the PR isn't being ignored, it's unreadable.

*This comment was created with AI assistance.*

src/main/java/org/apache/maven/plugin/compiler/DiagnosticLogger.java
desruisseaux: When two `case` are next to each other with no code between them, the intend is considered clear enough that the compiler does not emit any "fallthrough" warning. This is equivalent to `case MANDATORY_WARNING, WARNING -> (…some code…)` in the newer syntax.

desruisseaux:
This pull request has not been merged yet for the following reasons:

* It removes the `LOCALE` constant, while I would like to keep it. It is not because this constant is set to the `null` value that it is dead code. Replacing `null` by `Locale.JAPANESE` (for example) would have the expected effect: localize some messages to Japanese. Even if we are unlikely to change the `null` value to something else, the `LOCALE` constant is still useful for spotting easily the places where localization happens.
* It introduces a new configuration parameter (`messageLogType`) with a new enumeration (`MessageLogType`), which I think overlaps with the configuration of logging framework. Especially since Guillaume has done work on Maven logging recently, I think we should try to reftrofit this feature as configuration of the logging framework.
* It contains a `tryGetCompilerBundle()` which is very specific to JDK internal and which will be blocked by the JVM if Maven is modularized some day. I saw that Guillame has already done some modularisation experiments.

rmannibucau:
guess it is totally superseeded by recent changes so closing

Issue #1039, opened 2026-03-10
Dependencies of `<type>modular-processor</type>`  fail if project uses them with a different `type`


cbarlin:
### Affected version

4.0.0-rc-5

### Bug description

Not sure if this is a bug in the compiler plugin or in `javac`, but if you have a dependency that is marked with a type of `modular-processor` and it depends on a module that you also depend on (as in, type `modular-jar`), then `javac` throws an `java.lang.module.FindException`.

Using my (admittedly quite large, sorry) project as an example you can [see here](https://github.com/cbarlin/advanced-record-utils/blob/maven_4/aru-processor-core/pom.xml#L71) I've declared my dependencies like so:

```xml
<dependency>
    <groupId>io.avaje</groupId>
    <artifactId>avaje-inject</artifactId>
    <version>${avaje-inject.version}</version>
    <type>modular-jar</type>
</dependency>
<dependency>
    <groupId>io.avaje</groupId>
    <artifactId>avaje-inject-generator</artifactId>
    <version>${avaje-inject.version}</version>
    <type>modular-processor</type>
</dependency>
```

(note: `avaje-inject-generator` itself depends on `avaje-inject`).

When running `mvn compile` though, the below stack trace occurs:

<details>

<summary>Stack trace</summary>

```
[INFO] --- compiler:4.0.0-beta-4:compile (default-compile) @ aru-processor-core ---
[WARNING]  Parameter 'debug' (user property 'maven.compiler.debug') is deprecated: Setting this flag to false is replaced by <debuglevel>none</debuglevel>.
[INFO] Compiling all files.
[INFO] 
[ERROR] An exception has occurred in the compiler (25). Please file a bug against the Java compiler via the Java bug reporting page (https://bugreport.java.com) after checking the Bug Database (https://bugs.java.com) for duplicates. Include your program, the following diagnostic, and the parameters passed to the Java compiler in your report. Thank you.
java.lang.module.FindException: Module io.avaje.inject not found, required by io.avaje.inject.generator
        at java.base/java.lang.module.Resolver.findFail(Resolver.java:921)
        at java.base/java.lang.module.Resolver.resolve(Resolver.java:192)
        at java.base/java.lang.module.Resolver.bind(Resolver.java:315)
        at java.base/java.lang.module.Resolver.bind(Resolver.java:219)
        at java.base/java.lang.module.Configuration.resolveAndBind(Configuration.java:543)
[...]

PR #1042, opened 2026-03-15
test: add integration test for modular processor with shared dependencies

mariuszs:
Integration test for #1039.

The underlying issue (same GAV with different types resolved to a single path) is being fixed in Maven core via https://github.com/apache/maven/pull/11805. This PR adds a regression test only — no compiler plugin code changes.

### DIFF
--- a/src/it/modular-processor-shared-dep/annotation-user/pom.xml
+++ b/src/it/modular-processor-shared-dep/annotation-user/pom.xml
@@ -1,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one
+  ~ or more contributor license agreements.  See the NOTICE file
+  ~ distributed with this work for additional information
+  ~ regarding copyright ownership.  The ASF licenses this file
+  ~ to you under the Apache License, Version 2.0 (the
+  ~ "License"); you may not use this file except in compliance
+  ~ with the License.  You may obtain a copy of the License at
+  ~
+  ~   http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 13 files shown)


desruisseaux:
I think that the analysis is correct and this pull request seems to address the issue, but as a workaround. Ideally, it seems to me that the issue should be addressed in Maven core or in Maven resolver. It would allow to add only the dependencies that were really requested, instead of adding all dependencies that could potentially have been requested.

I was hopping that https://github.com/apache/maven/pull/11380 would help, but maybe it is only a first step. If I'm guessing correctly (I have not yet verified with a step-by-step debugging), the above-cited PR 11380 fixes the type of transitive dependencies when the type is only `modular-processor`, but we still have the issue that if a dependency is declared with both `modular-processor` and `modular-jar` types, only one of those types win while actually we need to keep both. @cstamas do yo think that this analysis is correct?

cstamas:
Yes, it is. Using _same GAV with different types is no-go currently_.

desruisseaux:
> Yes, it is. Using same GAV with different types is no-go currently.

So, it is correct to said that https://github.com/apache/maven/pull/11805 tries to make that possible?

desruisseaux:
For information, I verified that when using Maven 4.1.0-SNAPSHOT with the https://github.com/apache/maven/pull/11805 patch applied, the test provided in this pull request passes with not need to change anything in the compiler code (i.e. with `ToolExecutor` changes reverted).

desruisseaux:
Actually, instead of closing it, I was thinking to just discard the changes in `ExecutorTool` and keep the integration test.

mariuszs:
Done.

desruisseaux:
I pushed a `maven-4.1` branch. Can you edit this pull request for targeting that branch? This is because the master branch targets Maven 4.0.x, while this test requires Maven 4.1.

PR #1062, opened 2026-06-08
(docs) Add moditect-maven-plugin as it is a less complex way for Java 8 with java module descriptors

bmarwell:
Adds the less complex moditect plugin to the module-info docs which is probably more helpful than compiling sources twice.

---

- [X] Your pull request should address just one issue, without pulling in other changes.
- [ ] Run `mvn verify` to make sure basic checks pass.

### DIFF
--- a/src/site/markdown/examples/module-info.md
+++ b/src/site/markdown/examples/module-info.md
@@ -25,6 +25,8 @@ there is nothing special to do.
 
 # Java 8 projects with module-info
 
+## Using the maven-compiler-plugin
+
 Projects that want to be compatible with older versions of Java (i.e, 8 or below bytecode and API),
 but also want to provide a `module-info.java` for use on Java 9+ runtime,
 must be aware that they need to call `javac` twice:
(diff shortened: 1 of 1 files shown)


desruisseaux:
Actually the section saying that we need to run `javac` twice could be updated. It should not be needed any more since the addition of the multi-release support, although we would need to test for making sure. But in principle, a configuration like below should work with Java 8, with the `module-info.java` file in the `src/main/java_9` directory:

```xml
    <sources>
      <source>
        <directory>src/main/java</directory>
        <targetVersion>8</targetVersion>
      </source>
      <source>
        <directory>src/main/java_9</directory>
        <targetVersion>9</targetVersion>
      </source>
    </sources>
```

We could amend the pull request so that instead of being "two compilations versus one compilation", make it "explicit module-info versus module-info derived from the pom.xml". We may need to test the above snippet first for making sure that it works in this context too.

slachiewicz (approved):

desruisseaux (approved):
I tried the approach described in [above comment](https://github.com/apache/maven-compiler-plugin/pull/1062#issuecomment-4649195427) and it did not worked. It is a bit tricky to compile only a `module-info.java` file when everything else has been compiled as a class-path project. I just noticed that indeed, the existing text about using `maven-compiler-plugin` compiles the whole project twice. Therefore, I have no change to propose.

PR #1076, opened 2026-07-12
Fix incremental detection of empty sources, 4.x

### LINKED ISSUE #1000: Empty Java source files are always detected as changed by compiler:compile, which triggers full recompilation of all dependent modules.
### New feature, improvement proposal

Using maven-compiler-plugin:3.11.0 with the compile goal, empty Java source files are always detected as stale. This triggers unnecessary recompilation because the plugin treats these files as changed even when they are not.

In my case, the issue was caused by an accidental empty .java file. Removing the file resolved the problem, but the plugin should ideally ignore empty sources instead of marking them as stale.

wilx:
Class-based incremental compilation assumes every Java source produces an inferred output file. A zero-byte compilation unit produces no class file, so it was detected as modified on every invocation.

This change records whether a source is empty using the file attributes already collected during source scanning. A modified zero-byte source is treated as ignorable only when its inferred output does not exist. If an output does exist, the source remains stale so truncating an existing source is still detected as a change.

A regression test compiles an empty source twice with class-based incremental detection and verifies that the second invocation reports all classes as up to date.

Fixes #1000.

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/IncrementalBuild.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/IncrementalBuild.java
@@ -773,15 +773,12 @@ final class IncrementalBuild {
 
     /**
      * {@return whether the given list of modified files should not cause a recompilation}
-     * This method returns {@code true} if the given list is empty or contains only files
-     * with the {@link SourceFile#ignoreModification} set to {@code true}, or empty source
-     * files for which no output file exists.
+     * This method returns {@code true} if the given list is empty or if
+     * {@link SourceFile#isEmptyOrIgnorable()} returns {@code true} for every file.
      *
      * @param sourceFiles return value of {@link #getModifiedSources()}.
      */
     static boolean isEmptyOrIgnorable(List<SourceFile> sourceFiles) {
-        return sourceFiles.stream()
-                .allMatch(
-                        (s) -> s.ignoreModification || (s.isEmpty && Files.notExists(s.getOutputFile(), LINK_OPTIONS)));
+        return sourceFiles.stream().allMatch(SourceFile::isEmptyOrIgnorable);
     }
 }
--- a/src/main/java/org/apache/maven/plugin/compiler/IncrementalBuild.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/IncrementalBuild.java
@@ -774,11 +774,14 @@ final class IncrementalBuild {
     /**
      * {@return whether the given list of modified files should not cause a recompilation}
      * This method returns {@code true} if the given list is empty or contains only files
-     * with the {@link SourceFile#ignoreModification} set to {@code true}.
+     * with the {@link SourceFile#ignoreModification} set to {@code true}, or empty source
+     * files for which no output file exists.
      *
      * @param sourceFiles return value of {@link #getModifiedSources()}.
      */
     static boolean isEmptyOrIgnorable(List<SourceFile> sourceFiles) {
-        return sourceFiles.stream().allMatch((s) -> s.ignoreModification);
+        return sourceFiles.stream()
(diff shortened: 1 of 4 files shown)


wampiedriessen:
For this PullRequest, please also consider the case where a Java source file consists of only comments.
Making it effectively empty as well

wilx:
This would require parsing the Java file. There are parsers that can be used, I think. However, it seems like overkill for already marginal situation.

src/main/java/org/apache/maven/plugin/compiler/IncrementalBuild.java
```diff
@@ -774,11 +774,14 @@ List<SourceFile> getModifiedSources() {
     /**
      * {@return whether the given list of modified files should not cause a recompilation}
      * This method returns {@code true} if the given list is empty or contains only files
-     * with the {@link SourceFile#ignoreModification} set to {@code true}.
+     * with the {@link SourceFile#ignoreModification} set to {@code true}, or empty source
```
desruisseaux: It is a small detail, but I suggest to move "or empty source" on the next line. We try, when reasonable, to change lines at some natural sentence boundary (after a coma, or after sentence end, _etc_). In case of future changes, one sentence per line makes clearer in the `diff` which sentence changed (otherwise a change in one sentence appears as if 2 or 3 sentences were modified).

src/main/java/org/apache/maven/plugin/compiler/IncrementalBuild.java
```diff
@@ -774,11 +774,14 @@ List<SourceFile> getModifiedSources() {
...
+     * files for which no output file exists.
      *
      * @param sourceFiles return value of {@link #getModifiedSources()}.
      */
     static boolean isEmptyOrIgnorable(List<SourceFile> sourceFiles) {
-        return sourceFiles.stream().allMatch((s) -> s.ignoreModification);
+        return sourceFiles.stream()
+                .allMatch(
+                        (s) -> s.ignoreModification || (s.isEmpty && Files.notExists(s.getOutputFile(), LINK_OPTIONS)));
```
desruisseaux: Since `s` is referenced 3 times, it may be clearer to add an `isEmptyOrIgnorable()` method in `SourceFile` instead. I suggest to insert that method between the constructor and `isStandardOutputFile()`.

src/test/resources/unit/compiler-empty-source-change-detection-test/plugin-config.xml
```diff
@@ -0,0 +1,37 @@
...
+<project>
+  <build>
+    <directory>${project.basedir}/target</directory>
+    <outputDirectory>${project.basedir}/target/classes</outputDirectory>
+    <plugins>
+      <plugin>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <compileSourceRoots>
+            <compileSourceRoot>${project.basedir}/src/main/java</compileSourceRoot>
```
desruisseaux, line 29: Since this is the default value, is this declaration necessary?

wilx: It is needed. The test harness does not seem to provide the default value.

src/main/java/org/apache/maven/plugin/compiler/IncrementalBuild.java
wilx: Moved.

src/main/java/org/apache/maven/plugin/compiler/IncrementalBuild.java
wilx: Fixed.

desruisseaux:
Thanks!

PR #1086, opened 2026-07-20
[#1085] Document conservative incremental default on JDK < 23

### LINKED ISSUE #1085: Default incremental compilation recompiles the whole source set on any change (JDK < 23, no annotation processor)
### Summary

With the default incremental configuration and **no annotation processor** on the classpath, changing a single source file recompiles the **entire** source set instead of only the modified file, on **JDK < 23**. This contradicts the documented default and defeats incremental compilation for the common edit→compile loop on the JDK 17 / 21 LTS releases.

### Documented behaviour

`AbstractCompilerMojo#incrementalCompilation` javadoc, "Default value":

> If there is no annotation processor, then the default is `options,dependencies,sources`. It means that a full rebuild will be done if the compiler options or the dependencies changed, or if a source file has been deleted. **Otherwise, only the modified source files will be recompiled.**

### Actual behaviour

On JDK < 23 with `proc` unset and no annotation processor present, any single-file change logs

```
[INFO] Recompiling all files because at least one source file changed.
```

and recompiles the whole source set.

### Root cause

`AbstractCompilerMojo#hasAnnotationProcessor()` returns `true` unconditionally when `proc` is unset and the running JDK is older than 23 (because `javac` defaulted to `-proc:full` before Java 23). `amendincrementalCompilation()` then adds `rebuild-on-add,rebuild-on-change` to the default, forcing a full rebuild — even though **no** annotation processor is actually on the classpath.

On JDK 23+ the same method checks for real processor paths (`PROCESSOR_CLASSES` / `PROCESSOR_MODULES`) and correctly returns `false`, so the default behaves per-file.

### Reproduction

Minimal project: 3 independent classes, no dependencies, compiler-plugin default config. Full `compile`, change one class, `compile` again — count recompiled `.class` files:

| JDK | property | result |
|-----|----------|--------|
| 21  | *(default)* | recompiles all 3 |
| 21  | `-Dmaven.compiler.proc=none` | recompiles 1 |
| 21  | `-Dmaven.compiler.incrementalCompilation=sources` | recompiles 1 |
| 25  | *(default)* | recompiles 1 |

Note: on the released **Maven 4.0.0-rc-5** runtime the incremental state is not persisted at all, so every `compile` is a full rebuild regardless of these settings. This reproduces on the `maven-4.0.x` / `master` SNAPSHOT lines (i.e. rc-6+), tested against `maven-compiler-plugin` `master` (`4.0.0-beta-5-SNAPSHOT`).

### Expected fix

When `proc` is unset, gate the `hasAnnotationProcessor()` result on the presence of actual processor paths on **all** JDKs, not only JDK 23+. If no processor will run, a pre-23 `-proc:full` default still produces no generated files, so per-file incremental is safe — matching the documented default.

### Regression test

I have an invoker IT (`incremental-default-no-processor`: three independent classes, default config, modify one between two `compile` invocations, assert only one file is recompiled). It is **red on JDK < 23** and **green on JDK 23+** before the fix, and green on all JDKs after. Happy to open a PR with the IT plus the fix.

ascheman:
Fixes #1085.

Per @desruisseaux's steer on the issue: the full rebuild on any change (when `proc` is unset on JDK < 23) is intentional. Before Java 23 `javac` defaults to `-proc:full` and discovers annotation processors on the compile classpath, and recompiling everything avoids losing processor-generated state (e.g. missed `META-INF/services` registrations). So this keeps the behaviour and documents it instead of changing it.

**Changes**

- `AbstractCompilerMojo`: clarify the `incrementalCompilation` default-value javadoc — why, on JDK < 23 with `proc` unset, the plugin conservatively assumes a processor may be present, and that setting `proc` to `none` (or an explicit `incrementalCompilation`) restores per-file recompilation.
- New IT `incremental-proc-none-per-file`: with `proc=none`, changing one of three independent classes recompiles only that file. This exercises the documented workaround and is green on all JDKs (verified on 21 and 25).

Doc + test only; no behavioural change.

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
@@ -649,6 +649,16 @@ public abstract class AbstractCompilerMojo implements Mojo {
      * then the default value is same as above with the addition of {@code "rebuild-on-add,rebuild-on-change"}.
      * It means that a full rebuild will be done if any kind of change is detected.</p>
      *
+     * <p>Whether an annotation processor is considered present depends on the Java version when {@link #proc}
+     * is unset, because {@code javac} enables annotation processing by default before Java&nbsp;23
+     * ({@code -proc:full}) but disables it since Java&nbsp;23 ({@code -proc:none} unless a processor is
+     * configured). Consequently, on Java&nbsp;versions prior to&nbsp;23 the plugin conservatively assumes that
+     * a processor may be present — since {@code javac} would discover processors on the compile classpath —
+     * and therefore applies {@code "rebuild-on-add,rebuild-on-change"} by default, doing a full rebuild on any
+     * change even when no processor is actually present. Projects on Java&nbsp;&lt;&nbsp;23 that use no
+     * annotation processor can restore per-file recompilation by setting {@link #proc} to {@code "none"}
+     * (or by setting this {@code incrementalCompilation} property explicitly).</p>
+     *
      * @see #staleMillis
      * @see #fileExtensions
      * @see #showCompilationChanges
--- a/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
@@ -649,15 +649,15 @@ public abstract class AbstractCompilerMojo implements Mojo {
      * then the default value is same as above with the addition of {@code "rebuild-on-add,rebuild-on-change"}.
      * It means that a full rebuild will be done if any kind of change is detected.</p>
      *
-     * <p>Whether an annotation processor is considered present depends on the Java version when {@link #proc}
-     * is unset, because {@code javac} enables annotation processing by default before Java&nbsp;23
-     * ({@code -proc:full}) but disables it since Java&nbsp;23 ({@code -proc:none} unless a processor is
-     * configured). Consequently, on Java&nbsp;versions prior to&nbsp;23 the plugin conservatively assumes that
-     * a processor may be present — since {@code javac} would discover processors on the compile classpath —
-     * and therefore applies {@code "rebuild-on-add,rebuild-on-change"} by default, doing a full rebuild on any
-     * change even when no processor is actually present. Projects on Java&nbsp;&lt;&nbsp;23 that use no
-     * annotation processor can restore per-file recompilation by setting {@link #proc} to {@code "none"}
-     * (or by setting this {@code incrementalCompilation} property explicitly).</p>
+     * <p>Whether an annotation processor is considered present depends on the Java version when {@link #proc} is unset,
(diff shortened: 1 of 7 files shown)


src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -649,6 +649,16 @@ final Charset charset() {
      * then the default value is same as above with the addition of {@code "rebuild-on-add,rebuild-on-change"}.
      * It means that a full rebuild will be done if any kind of change is detected.</p>
      *
+     * <p>Whether an annotation processor is considered present depends on the Java version when {@link #proc}
```
desruisseaux: This is a minor detail, but when reasonable I like to keep each sentence, or part of a sentence delimited by coma, per line. In this case, it would move _"is unset,"_ on the previous line. The intent is to make clearer `diff` if we modify the text in the future. When new lines happen at random locations, a change in one sentence appears as if 2 or 3 sentences where potentially impacted. When new lines happen at natural boundaries, `diff` shows only the sentences (or sentence parts delimited by coma) that were really modified.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -649,6 +649,16 @@ final Charset charset() {
...
+     * is unset, because {@code javac} enables annotation processing by default before Java&nbsp;23
+     * ({@code -proc:full}) but disables it since Java&nbsp;23 ({@code -proc:none} unless a processor is
+     * configured). Consequently, on Java&nbsp;versions prior to&nbsp;23 the plugin conservatively assumes that
```
desruisseaux: Do we really need no break space (`&nbsp;`) before "Java" and "versions" here?

desruisseaux (approved):

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
ascheman: Done in f84903d — reflowed the paragraph so each line breaks at a sentence or comma-delimited clause, and `is unset,` now sits with its clause on the previous line. Agreed on the cleaner diffs, thanks.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
ascheman: No, we do not — removed them in f84903d, plain spaces now. They were over-cautious on my part.

desruisseaux: Thanks. Note (no need for a change): I was not suggesting to remove all of them, I see a value for no-break space between "Java" and "23". I was thinking mostly to the case between ordinary words, such as "Java versions".

No need to change again, it was just a note.

PR #1088, opened 2026-07-27
Remove deprecated setCompilerVersion() call

elharo:
`CompilerConfiguration.setCompilerVersion()` is deprecated and no longer evaluated by the underlying compilers. The `compilerVersion` field in `AbstractCompilerMojo` is already `@Deprecated` with this Javadoc:

> This parameter is no longer evaluated by the underlying compilers, instead the actual version of the `javac` binary is automatically retrieved.

This removes the no-op call to eliminate the deprecation warning while keeping the `@Deprecated` field for backward compatibility — users who still set `maven.compiler.compilerVersion` will get a deprecation warning but not a build error."


ascheman (changes requested):
Thanks for chasing down these deprecation no-ops. The code change itself is correct: `CompilerConfiguration.setCompilerVersion()` is a deprecated no-op (plexus-compiler no longer evaluates it — a forked `javac`'s version is auto-detected via `<executable>`), and this was the only place in `src/main` that read the `compilerVersion` parameter. So the removal is behavior-preserving, and keeping the `@Deprecated` field for backward compatibility is the right call.

One thing to address before this goes in — because this PR is what makes it wrong:

With this call gone, `compilerVersion` is now fully inert (nothing reads it), yet the docs still present it as a working feature. In `src/site/markdown/examples/compile-using-different-jdk.md.vm`:

> The `compilerVersion` parameter can be used to specify the version of the compiler that the plugin will use. However, you also need to set `fork` to `true` for this to work.
> ```xml
> <fork>true</fork>
> <executable></executable>
> <compilerVersion>1.3</compilerVersion>
> ```

After this PR that `<compilerVersion>1.3</compilerVersion>` line does nothing. The `<fork>` + `<executable>` mechanism still selects the JDK and is fine to keep — it's just the `<compilerVersion>` element (and the sentence describing it) that's now dead. Could you drop those from the example in this same PR, so we don't finish deprecating the parameter while the docs still advertise it? Happy to approve once that's in.

Issue #333, opened 2025-06-01
Compiler is too chatty, but on top says wrong things


cstamas:
### Affected version

3.14.0

### Bug description

On Maven master this is spit out by plugin before it goes for longest compilation time out of all modules:
```
[INFO] --- compiler:3.14.0:compile (default-compile) @ maven-support ---
[INFO] Recompiling the module because of changed dependency.
[INFO] Compiling 19 source files with javac [debug deprecation release 17] to target/classes
```

This is too much and wrong:
* I don't care why it recompiles (as info is repeated for each module), first line should be DEBUG
* second line is too much, and wrong: if it compiles only 19 source files (it does not), why it takes so long
* I am not interested where it compiles
* IMO plugin should emit one line like `Compiling source files with javac [debug deorecation release=17]`

PS: maven-support module really has 19 source files, but a TON of generated sources as well (biggest in whole reactor).

desruisseaux:
The chatty part should be easy to resolve, but the part about the wrong count may be more problematic (unless we just remove that line) because the incremental compilation on the 3.x branch is known to have issues such as [MCOMPILER-209](https://issues.apache.org/jira/browse/MCOMPILER-209). Since incremental compilation has been reworked in the 4.x version, do we still need to fix the incremental compilation of 3.x? Or do we just remove those lines on 3.x?

zhaoyunxing92:
I tried it (maven 4.0.0-rc-4) and it is indeed much faster. This problem that has been around for years should be solved. That's cool.

zhaoyunxing92:
If you just modified the file adding method and did not involve import, you can use `-Dmaven.compiler.useIncrementalCompilation=false`, which will make the second compilation faster (without clean). In short, we look forward to the next version, which improves this code.
```shell
mvn package -Dmaven.compiler.useIncrementalCompilation=false
```

PR #1089, opened 2026-07-27
Demote recompiling message from INFO to DEBUG (#333)

elharo:
Part of  #333

Changes the 'Recompiling all files because ...' log message from INFO to DEBUG level. This message is emitted for every module during incremental builds and is noisy at the default log level.

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/ToolExecutor.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/ToolExecutor.java
@@ -389,7 +389,7 @@ public class ToolExecutor {
             }
             if (causeOfRebuild != null) {
                 if (!sourceFiles.isEmpty()) { // Avoid misleading message such as "all sources changed".
-                    logger.info(causeOfRebuild);
+                    logger.debug(causeOfRebuild);
                 }
             } else {
                 isPartialBuild = true;


ascheman (approved):
LGTM — demoting the per-module "Recompiling all files…" message to DEBUG is a sensible noise reduction for the default log level, and it's guarded by the existing `!sourceFiles.isEmpty()` check so it won't fire on empty sets. Thanks @elharo.

desruisseaux:
The comment in the review saying that we should probably keep the `info` level when the user explicitly requested that information (with the `showCompilationChanges` configuration option) has not been answered. There is also the issue that we currently flood the debug level with so many records that logging anything at this level is practically lost, because `debug` is often used as a `trace` level providing no way for users to debug their build configuration without being drowned in Maven internal details.

I have create pull request #1108 for the `showCompilationChanges` flag. About the flood at debug level, I have no solution for now, as this is wider problem that the compiler plugin. A solution may be to create a `Log.trace` method and demote many `debug` levels to `trace` among all Maven core and plugins.

ascheman:
> I have create pull request #1108 for the `showCompilationChanges` flag. About the flood at debug level, I have no solution for now, as this is wider problem that the compiler plugin. A solution may be to create a `Log.trace` method and demote many `debug` levels to `trace` among all Maven core and plugins.

I like this idea (we have discussed and applied it elsewhere already).

PR #1090, opened 2026-07-28
[MCOMPILER-595] Enable parameter metadata by default

### LINKED ISSUE #858: [MCOMPILER-595] Make -parameters enabled by default
**[Tamas Cservenak](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=cstamas)** opened **[MCOMPILER-595](https://issues.apache.org/jira/browse/MCOMPILER-595?redirect=false)** and commented

More and more frameworks rely on reflective access to parameter names.

---

2 votes, 2 watchers

wilx:
## Summary

- Enable `-parameters` by default for both `compiler:compile` and `compiler:testCompile`.
- Document that `maven.compiler.parameters=false` opts out of parameter-name metadata.
- Extend the MCOMPILER-298 integration test to verify both the new default and the explicit opt-out.

## Motivation

More frameworks rely on reflection to obtain method parameter names. Generating that metadata by default makes those names available without requiring every project to opt in explicitly, while retaining a configuration switch for projects that do not want the metadata.

Fixes https://github.com/apache/maven-compiler-plugin/issues/858

## Validation

`JAVA_HOME=/opt/jdks/latest-21 /opt/maven/latest-4/bin/mvn -ntp -Prun-its -Dinvoker.test=MCOMPILER-298 clean verify`

### DIFF
--- a/src/it/MCOMPILER-298/invoker.properties
+++ b/src/it/MCOMPILER-298/invoker.properties
@@ -16,3 +16,4 @@
 # under the License.
 
 invoker.goals = test
+invoker.goals.2 = clean test -Dmaven.compiler.parameters=false -DexpectedParameterNamePresent=false
--- a/src/it/MCOMPILER-298/pom.xml
+++ b/src/it/MCOMPILER-298/pom.xml
@@ -45,9 +45,6 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
         <version>@project.version@</version>
-        <configuration>
-          <parameters>true</parameters>
-        </configuration>
       </plugin>
     </plugins>
   </build>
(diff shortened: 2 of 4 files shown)


desruisseaux:
In Maven 3, the Maven Compiler Plugin often overwrote the default options of the standard `javac` command with Maven's own opinion of what these defaults should be. The consequence was, for example, arbitrary and non-obvious default values for the `-target` option.

In the 4.x line of Maven Compiler Plugin, a design goal so far was to have the same defaults as the standard `javac` command. A `javac` option is not added unless explicitly requested, in an attempt to be less surprising for developers who read the Oracle `javac` man page. This pull request would be a first breach in this attempt to be consistent.

Given that the default value of this parameter can be set by the `maven.compiler.parameters` property, would it be okay for projects who want such default to specify it in a parent `pom.xml`?

wilx:
While I have to enable it all the time for Spring Framework, I can deal with it. I saw the https://github.com/apache/maven-compiler-plugin/issues/858 and ran with it. I guess both the PR and the issue should be closed then.

desruisseaux:
Thanks. I'm closing this pull request on the basis of consistency with default behaviour of standard `javac`. I re-targeted #858 to 3.x since the 3.x line of the Compiler Plugin already has defaults that diverge from `javac` defaults. Therefore, maybe the question of whether to provide this default is still open for 3.x.

PR #1092, opened 2026-07-28
[MCOMPILER-569] Document implicitly compiled excluded sources

### LINKED ISSUE #778: [MCOMPILER-569] Plugin doesn't exclude module-info
**[Pavel_K](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=pavel_k)** opened **[MCOMPILER-569](https://issues.apache.org/jira/browse/MCOMPILER-569?redirect=false)** and commented

The following code must exclude maven-info file.

```java
<plugin>
     <version>3.11.0</version>
     <configuration>
         <source>${maven.compiler.source}</source>
         <target>${maven.compiler.target}</target>
         <excludes>
             <exclude>**/module-info.java</exclude>
         </excludes>
     </configuration>
 </plugin>
```

However, plugin compiles module-info and adds module-info.class to jar file

---

**Affects:** 3.11.0

wilx:
Clarify the `compile` and `testCompile` exclusion parameter documentation: an excluded source such as `module-info.java` may still be compiled when `javac` finds it implicitly. Link both parameters to `<implicit>`, which can be set to `none` to suppress class generation for implicitly referenced sources.

This documents the distinction between the plugin's explicit source filtering and `javac`'s implicit source discovery.

Fixes #778

       pulling in other changes.
       commit message.

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/CompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/CompilerMojo.java
@@ -109,6 +109,10 @@ public class CompilerMojo extends AbstractCompilerMojo {
 
     /**
      * A list of exclusion filters for the compiler.
+     * A file such as {@code module-info.java} may still be compiled if it is implicitly referenced by the compiler,
+     * even if it has been excluded.
+     *
+     * @see AbstractCompilerMojo#implicit
      */
     @Parameter
     private Set<String> excludes = new HashSet<>();
--- a/src/main/java/org/apache/maven/plugin/compiler/CompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/CompilerMojo.java
@@ -109,8 +109,9 @@ public class CompilerMojo extends AbstractCompilerMojo {
 
     /**
      * A list of exclusion filters for the compiler.
-     * A file such as {@code module-info.java} may still be compiled if it is implicitly referenced by the compiler,
-     * even if it has been excluded.
+     * Excluding a source file only prevents the plugin from passing it explicitly to the compiler.
+     * The compiler may still find it on the source path and generate a class file for it.
+     * This can happen with {@code module-info.java}.
      *
      * @see AbstractCompilerMojo#implicit
      */
--- a/src/main/java/org/apache/maven/plugin/compiler/TestCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/TestCompilerMojo.java
@@ -96,8 +96,9 @@ public class TestCompilerMojo extends AbstractCompilerMojo {
 
     /**
      * A list of exclusion filters for the compiler.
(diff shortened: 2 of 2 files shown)


src/main/java/org/apache/maven/plugin/compiler/CompilerMojo.java
```diff
@@ -109,6 +109,10 @@ public class CompilerMojo extends AbstractCompilerMojo {
 
     /**
      * A list of exclusion filters for the compiler.
+     * A file such as {@code module-info.java} may still be compiled if it is implicitly referenced by the compiler,
```
elharo: I don't understand what "implicitly referenced by the compiler" means.

elharo (dismissed):

src/main/java/org/apache/maven/plugin/compiler/CompilerMojo.java
wilx: This is described in <https://docs.oracle.com/en/java/javase/17/docs/specs/man/javac.html#searching-for-module-package-and-type-declarations>. The relevant lines are 

> When compiling code for a module, the compiler also needs to have available the declaration of that module.
> A successful search may produce a class file, a source file, or both. If both are found, then you can use the [-Xprefer](https://docs.oracle.com/en/java/javase/17/docs/specs/man/javac.html#option-Xprefer) option to instruct the compiler which to use.
> If a search finds and uses a source file, then by default javac compiles that source file. This behavior can be altered with [-implicit](https://docs.oracle.com/en/java/javase/17/docs/specs/man/javac.html#option-implicit).

Feel free to suggest better wording.

wilx:
I have pushed a change with different wording.

desruisseaux (approved):

src/main/java/org/apache/maven/plugin/compiler/CompilerMojo.java
```diff
@@ -109,6 +109,10 @@ public class CompilerMojo extends AbstractCompilerMojo {
 
     /**
      * A list of exclusion filters for the compiler.
+     * Excluding a source file only prevents the plugin from passing it explicitly to the compiler. The compiler may
+     * still find it on the source path and generate a class file for it. This can happen with {@code module-info.java}.
```
desruisseaux: This is a minor detail, but at least on `master` I tend to change line after a dot or coma when possible. Therefore the comment could be (exact same text):

```java
     * Excluding a source file only prevents the plugin from passing it explicitly to the compiler.
     * The compiler may still find it on the source path and generate a class file for it.
     * This can happen with {@code module-info.java}.
```

The reason for this formatting is that, if we edit the sentences again in the future, `git diff` shows only the sentence that are impacted. Otherwise we have a mix of potentially 3 sentences that are impacted and it is not immediately obvious that the sentence just before or just after is actually unchanged.

wilx: Fixed.

PR #1097, opened 2026-07-29
Remove version numbers of plugins inherited from parent pom.xml

desruisseaux:
The plugins must be upgraded for compatibility with Java 26. But we do not need any more to declare those version numbers ourselves. It is now sufficient to inherit them from the parent `pom.xml`.

This pull request opportunistically changes the `invoker.toolchain.jdk.version` property of an integration test to match the `pom.xml`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -96,12 +96,8 @@ under the License.
 
     <project.build.outputTimestamp>2026-01-27T00:46:35Z</project.build.outputTimestamp>
 
-    <!-- TODO remove with next parent -->
-    <maven4x.site.path>plugins-archives/${project.artifactId}-LATEST-4.x</maven4x.site.path>
-    <version.maven-compiler-plugin>3.14.1</version.maven-compiler-plugin>
-    <version.maven-dependency-plugin>3.9.0</version.maven-dependency-plugin>
-    <version.maven-invoker-plugin>3.9.1</version.maven-invoker-plugin>
     <!-- 4.x publish -->
+    <maven4x.site.path>plugins-archives/${project.artifactId}-LATEST-4.x</maven4x.site.path>
     <maven.site.path>${maven4x.site.path}</maven.site.path>
 
     <maven.compiler.source />


src/it/multirelease-patterns/singleproject-toolchains/invoker.properties
```diff
@@ -15,4 +15,4 @@
 # specific language governing permissions and limitations
 # under the License.
 invoker.goals = verify -Pmulti-jar
-invoker.toolchain.jdk.version=1.9
+invoker.toolchain.jdk.version=1.8
```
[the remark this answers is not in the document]
desruisseaux: @elharo : the problem is that I don't know if this property means to run the test with the specified Java version, or if the specified Java version must be available. The `pom.xml` seems to expect a JDK 8 to be available.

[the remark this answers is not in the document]
desruisseaux: Ok, I removed the `invoker.toolchain.jdk.version` change. I verified that no matter if we set the Java version to 8 ot 9, the test is skipped anyway, both locally on my machine and in GitHub action.

As it stands today, the `it/multirelease-patterns/singleproject-toolchains` test is useless and should be either removed or changed into something which is actually executed.

slawekjaranowski (approved):

pom.xml
slawekjaranowski, line 105: We also should use `javaVersion` here 
https://maven.apache.org/pom/asf/

pom.xml
```diff
@@ -96,12 +96,8 @@ under the License.
...
     <project.build.outputTimestamp>2026-01-27T00:46:35Z</project.build.outputTimestamp>
 
-    <!-- TODO remove with next parent -->
-    <maven4x.site.path>plugins-archives/${project.artifactId}-LATEST-4.x</maven4x.site.path>
-    <version.maven-compiler-plugin>3.14.1</version.maven-compiler-plugin>
-    <version.maven-dependency-plugin>3.9.0</version.maven-dependency-plugin>
-    <version.maven-invoker-plugin>3.9.1</version.maven-invoker-plugin>
     <!-- 4.x publish -->
+    <maven4x.site.path>plugins-archives/${project.artifactId}-LATEST-4.x</maven4x.site.path>
     <maven.site.path>${maven4x.site.path}</maven.site.path>
```
slawekjaranowski, line 102: this can be changed to: `maven.site.path.suffix`
https://maven.apache.org/pom/maven/#site-publication

desruisseaux: Would the following be a replacement for both `<maven4x.site.pat>` and `<maven.site.path>`? In particular, would it include the `plugins-archives` part ?

```xml
<maven.site.path.suffix>LATEST-4.x</maven.site.path.suffix>
```

desruisseaux: Will apply the proposed changes in a separated commit. In particular, I would need to test the suffix one.

pom.xml
desruisseaux: Will apply in a future commit, I need a little bit of time for testing.

PR #1098, opened 2026-07-29
Remove `src/it/multirelease-patterns/singleproject-toolchains`.

desruisseaux:
I propose to remove this integration test, because it cannot be executed without specific Java versions installed. In practice, it is rarely executed either locally or by GitHub actions. Even if we managed to get it working on GitHub actions, it would be only temporary and would quickly become outdated as we upgrade our runners with more recent Java versions.


elharo (commented):
It's tricky but maybe there's some other way to get it to work? This is important functionality that should be tested.

desruisseaux:
We would need to find a way to keep the test `pom.xml` consistent with GitHub Action (GHA) configuration. I presume that this is possible, but I'm not familiar enough with Toolchains configuration in GHA, and how to propagate that information to the `pom.xml`, for doing that on my side…

slawekjaranowski:
Agree with @elharo should be tested somehow ...

But also agree with @desruisseaux this test is orphan and not working now ....

one we need to use a ranges for toolchains like for property for invoker:

```
invoker.toolchain.jdk.version=[9,)
```

slawekjaranowski (approved):
please create an issue for preparing a new tests for toolchains

desruisseaux:
Okay, I will close this pull request and create an issue instead.

desruisseaux:
Replaced by #1099.

PR #1102, opened 2026-07-31
Recompile when dependencies change

### LINKED ISSUE #1087: No recompilation if the dependencies have changed
### Affected version

3.15.0

### Bug description

To reproduce:
* unzip [sample.zip](https://github.com/user-attachments/files/30305357/sample.zip)
* run `mvn compile`
* delete the `dependencies` section from the POM
* run `mvn compile`

Expected: the last step fails because the dependencies have changed and thus recompilation is triggered and thus compilation fails due to missing classes

Actual: the last step succeeds and the compiler plugin says "Nothing to compile - all classes are up to date."

wilx:
## Summary

- persist the ordered class path and module path in the plugin-local `dependencies.lst` state file
- collect persisted metadata and the current-build timestamp fallback in one cached scan per normalized dependency path, removing the previous duplicate traversal
- record regular dependency files as size and modification time, and exploded dependency directories as a stable metadata fingerprint
- trigger the existing `changed dependency` rebuild reason when dependencies are added, removed, reordered, relocated, or replaced at the same path
- add unit coverage for the consolidated detector and integration coverage for dependency removal and same-path SNAPSHOT replacement

## Why

The 3.x incremental check only considered dependency timestamps relative to the current Maven session. It could therefore miss a dependency removed between invocations or a JAR replaced at the same path before the next invocation, leaving stale application bytecode.

The persisted comparison and the existing session-relative fallback are now consolidated. A missing baseline still detects dependencies rebuilt during the current Maven invocation, while unchanged builds no longer walk exploded dependencies once for persisted state and again for the timestamp fallback.

## Validation

- `JAVA_HOME=/opt/jdks/latest-8 /opt/maven/latest/bin/mvn clean verify` — 36 tests passed
- `JAVA_HOME=/opt/jdks/latest-21 /opt/maven/latest/bin/mvn clean verify` — 36 tests passed
- focused MCOMPILER-349, MCOMPILER-474, dependency-removal, and same-path-replacement integration tests passed
- `JAVA_HOME=/opt/jdks/latest-21 /opt/maven/latest/bin/mvn -Prun-its clean verify` — 78 builds passed, 5 environment-specific skips

Fixes #1087

---

- [x] An issue is filed for this change: #1087. This pull request addresses only that issue.
- [x] Each commit in the pull request has a meaningful subject line and body.
- [x] The pull request title describes the issue. The template's `[MCOMPILER-XXX]` convention is not applicable because this is tracked as GitHub issue #1087.
- [x] The pull request description explains what the pull request does, how, and why.
- [x] Ran `mvn clean verify`.
- [x] Ran the integration tests successfully with `mvn -Prun-its clean verify`.

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
@@ -80,7 +80,6 @@ import org.codehaus.plexus.compiler.util.scan.mapping.SourceMapping;
 import org.codehaus.plexus.compiler.util.scan.mapping.SuffixMapping;
 import org.codehaus.plexus.languages.java.jpms.JavaModuleDescriptor;
 import org.codehaus.plexus.languages.java.version.JavaVersion;
-import org.codehaus.plexus.util.FileUtils;
 import org.eclipse.aether.RepositorySystem;
 import org.eclipse.aether.artifact.Artifact;
 import org.eclipse.aether.artifact.ArtifactTypeRegistry;
@@ -981,7 +980,19 @@ public abstract class AbstractCompilerMojo extends AbstractMojo {
                                 && !canUpdateTarget)
                         ? "immutable single output file"
                         : null;
-                String dependencyChanged = isDependencyChanged() ? "changed dependency" : null;
+                String dependencyChanged = DependencyState.hasChanged(
+                                incrementalBuildHelper,
+                                getOutputDirectory(),
+                                getClasspathElements(),
+                                getModulepathElements(),
+                                new DependencyState.Configuration(
+                                        fileExtensions,
+                                        getBuildStartTimeInstant().orElse(null),
+                                        staleMillis,
+                                        getLog(),
+                                        showCompilationChanges))
+                        ? "changed dependency"
+                        : null;
                 String sourceChanged = isSourceChanged(compilerConfiguration, compiler) ? "changed source code" : null;
                 String inputFileTreeChanged = hasInputFileTreeChanged(incrementalBuildHelper, sources)
                         ? "added or removed source files"
@@ -1733,81 +1744,6 @@ public abstract class AbstractCompilerMojo extends AbstractMojo {
                     .toArray(String[]::new);
         }
         return processors == null || processors.length == 0 ? null : processors;
-    }
-
-    /**
-     * We just compare the timestamps of all local dependency files (inter-module dependency classpath) and the own
-     * generated classes and if we got a file which is &gt;= the build-started timestamp, then we caught a file which
-     * got changed during this build.
-     *
-     * @return {@code true} if at least one single dependency has changed.
-     */
-    protected boolean isDependencyChanged() {
-        final Instant buildStartTime = getBuildStartTimeInstant().orElse(null);
-        if (buildStartTime == null) {
-            // we just cannot determine it, so don't do anything beside logging
-            getLog().debug("Cannot determine build start time, skipping incremental build detection.");
-            return false;
-        }
-
-        if (fileExtensions == null || fileExtensions.isEmpty()) {
-            fileExtensions = new HashSet<>(Arrays.asList("class", "jar"));
-        }
-
-        List<String> pathElements = new ArrayList<>();
-        pathElements.addAll(getClasspathElements());
-        pathElements.addAll(getModulepathElements());
-
-        for (String pathElement : pathElements) {
-            Path artifactPath = Paths.get(pathElement);
-
-            // Search files only on dependencies (other modules), not on the current project,
-            if (Files.isDirectory(artifactPath)
-                    && !artifactPath.equals(getOutputDirectory().toPath())) {
-                try (Stream<Path> walk = Files.walk(artifactPath)) {
-                    if (walk.anyMatch(p -> hasNewFile(p, buildStartTime))) {
-                        return true;
-                    }
-                } catch (IOException ex) {
-                    // we just cannot determine it, so don't do anything beside logging
-                    getLog().warn("I/O error walking the path: " + ex.getMessage());
-                    return false;
-                }
-            } else if (hasNewFile(artifactPath, buildStartTime)) {
-                return true;
-            }
-        }
-
-        // obviously there was no new file detected.
-        return false;
-    }
-
-    /**
-     * @param file entry to check
-     * @param buildStartTime time build start
-     * @return if any changes occurred
-     */
-    private boolean hasNewFile(Path file, Instant buildStartTime) {
-        if (Files.isRegularFile(file)
-                && fileExtensions.contains(
-                        FileUtils.extension(file.getFileName().toString()))) {
-            try {
-                Instant lastModifiedTime = Files.getLastModifiedTime(file)
-                        .toInstant()
-                        .minusMillis(staleMillis)
-                        .truncatedTo(ChronoUnit.MILLIS);
-                boolean hasChanged = lastModifiedTime.isAfter(buildStartTime);
-                if (hasChanged && (getLog().isDebugEnabled() || showCompilationChanges)) {
-                    getLog().info("\tNew dependency detected: " + file.toAbsolutePath());
-                }
-                return hasChanged;
-            } catch (IOException ex) {
-                // we just cannot determine it, so don't do anything beside logging
-                getLog().warn("I/O error reading the lastModifiedTime: " + ex.getMessage());
-            }
-        }
-
-        return false;
     }
 
     private List<String> resolveProcessorPathEntries() throws MojoExecutionException {
--- a/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
@@ -980,19 +980,20 @@ public abstract class AbstractCompilerMojo extends AbstractMojo {
                                 && !canUpdateTarget)
                         ? "immutable single output file"
                         : null;
-                String dependencyChanged = DependencyState.hasChanged(
-                                incrementalBuildHelper,
-                                getOutputDirectory(),
-                                getClasspathElements(),
-                                getModulepathElements(),
-                                new DependencyState.Configuration(
-                                        fileExtensions,
-                                        getBuildStartTimeInstant().orElse(null),
-                                        staleMillis,
-                                        getLog(),
-                                        showCompilationChanges))
-                        ? "changed dependency"
-                        : null;
+                String dependencyChanged = null;
+                if (DependencyState.hasChanged(
+                        incrementalBuildHelper,
+                        getOutputDirectory(),
+                        getClasspathElements(),
+                        getModulepathElements(),
+                        new DependencyState.Configuration(
+                                fileExtensions,
+                                getBuildStartTimeInstant().orElse(null),
+                                staleMillis,
+                                getLog(),
+                                showCompilationChanges))) {
+                    dependencyChanged = "changed dependency";
+                }
                 String sourceChanged = isSourceChanged(compilerConfiguration, compiler) ? "changed source code" : null;
                 String inputFileTreeChanged = hasInputFileTreeChanged(incrementalBuildHelper, sources)
                         ? "added or removed source files"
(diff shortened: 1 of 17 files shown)


src/main/java/org/apache/maven/plugin/compiler/DependencyState.java
```diff
@@ -0,0 +1,306 @@
...
+
+        if (hasPreviousState && !oldState.equals(newState)) {
+            if (configuration.log.isDebugEnabled() || configuration.showChanges) {
+                logChanges(oldState, newState, configuration.log);
+            }
+            return true;
+        }
+        if (changedDependency != null) {
+            if (configuration.log.isDebugEnabled() || configuration.showChanges) {
+                configuration.log.info("\tNew dependency detected: " + changedDependency.toAbsolutePath());
```
elharo, line 107: I wouldn't log this above debug level

src/main/java/org/apache/maven/plugin/compiler/DependencyState.java
```diff
@@ -0,0 +1,306 @@
...
+                        ? dependency
+                        : null;
+                return new ModificationState(fileMetadata(attributes), changedDependency);
+            } catch (IOException e) {
+                context.log.warn("I/O error reading dependency state: " + dependency + ": " + e.getMessage());
+                return new ModificationState("unreadable", null);
+            }
+        }
+
+        List<Path> files;
```
elharo: This belongs inside the try, as does all the code that depends on it. Don't split variable declaration and initialization

src/main/java/org/apache/maven/plugin/compiler/DependencyState.java
```diff
@@ -0,0 +1,306 @@
...
+        for (Path file : files) {
+            try {
+                BasicFileAttributes attributes = readAttributes(file);
+                update(digest, dependency.relativize(file).toString());
+                update(digest, fileMetadata(attributes));
+                if (changedDependency == null && changedSinceBuildStart(attributes, context)) {
+                    changedDependency = file;
+                }
+            } catch (IOException e) {
+                if (!unreadable) {
```
elharo: should you just break out of the loop in this case?

src/main/java/org/apache/maven/plugin/compiler/DependencyState.java
```diff
@@ -0,0 +1,306 @@
...
+        String recordedState = unreadable ? "unreadable" : files.size() + ":" + toHexString(digest.digest());
+        return new ModificationState(recordedState, changedDependency);
+    }
+
+    private static BasicFileAttributes readAttributes(Path file) throws IOException {
+        return Files.readAttributes(file, BasicFileAttributes.class);
+    }
+
+    private static String fileMetadata(BasicFileAttributes attributes) {
+        return attributes.size() + ":" + attributes.lastModifiedTime().toMillis();
```
elharo, line 185: Seems wonky to be using a string where this method is called instead of the raw attributes.

src/main/java/org/apache/maven/plugin/compiler/DependencyState.java
```diff
@@ -0,0 +1,306 @@
...
+            throw new IllegalStateException(e);
+        }
+    }
+
+    private static void update(MessageDigest digest, String value) {
+        digest.update(value.getBytes(StandardCharsets.UTF_8));
+        digest.update((byte) 0);
+    }
+
+    private static String toHexString(byte[] bytes) {
```
elharo, line 211: This method isn't very risky, but I wouldn't be surprised if there's a utility method to do this somewhere in the classpath already, and if so we should use it. I think there's one in the JDK circa Java 17 but we might not depend on that version yet.

src/main/java/org/apache/maven/plugin/compiler/DependencyState.java
```diff
@@ -0,0 +1,306 @@
...
+                        .lastModifiedTime()
+                        .toInstant()
+                        .minusMillis(context.staleMillis)
+                        .truncatedTo(ChronoUnit.MILLIS)
+                        .isAfter(context.buildStartTime);
+    }
+
+    private static MessageDigest newDigest() {
+        try {
+            return MessageDigest.getInstance("SHA-256");
```
elharo, line 200: unusual choice. Most of the time it would be SHA-1 for speed or SHA-512 for security. I'm not sure securtiy matters here. Does it?

src/main/java/org/apache/maven/plugin/compiler/DependencyState.java
```diff
@@ -0,0 +1,306 @@
...
+
+    private static void logChanges(List<String> oldState, List<String> newState, Log log) {
+        List<String> oldPaths = paths(oldState);
+        List<String> newPaths = paths(newState);
+        DeltaList<String> pathChanges = new DeltaList<>(oldPaths, newPaths);
+        for (String dependencyAdded : pathChanges.getAdded()) {
+            log.info("\tDependency path (+): " + dependencyAdded);
+        }
+        for (String dependencyRemoved : pathChanges.getRemoved()) {
+            log.info("\tDependency path (-): " + dependencyRemoved);
```
elharo, line 231: feels like this should all be debug level

src/main/java/org/apache/maven/plugin/compiler/DependencyState.java
```diff
@@ -0,0 +1,306 @@
...
+            try {
+                Files.write(stateFile, newState);
+            } catch (IOException e) {
+                configuration.log.warn("Error while writing new dependency status: " + stateFile);
+            }
+        }
+
+        if (hasPreviousState && !oldState.equals(newState)) {
+            if (configuration.log.isDebugEnabled() || configuration.showChanges) {
+                logChanges(oldState, newState, configuration.log);
```
elharo, line 101: checking debug here is good, but it's info level in the logChanges method. logChanges should log at debug level

elharo (changes requested):

src/main/java/org/apache/maven/plugin/compiler/DependencyState.java
wilx: The `log.info` is there because it is possible to get the line printed with user supplied `maven.compiler.showCompilationChanges` property. If it were `log.debug` it would not print if the user asked via the property. I can change it to `debug` for `configuration.log.isDebugEnabled()` and keep the `info` for `configuration.showChanges`. But that seems unnecessarily nuanced.

src/main/java/org/apache/maven/plugin/compiler/DependencyState.java
wilx: It fits into how it is stored and used. This way the checksum of files in a dir is only computed via checksum of their metadata checksums/strings, there is only one `update` function. Doing anything else would complicate things unnecessarily.

src/main/java/org/apache/maven/plugin/compiler/DependencyState.java
wilx: Why unusual? It seems to be used a lot everywhere. It is AFAIK what Git has chosen as successor of SHA-1. It is a middle ground between the speed and security.

src/main/java/org/apache/maven/plugin/compiler/DependencyState.java
elharo: I only see one call of this method?

src/main/java/org/apache/maven/plugin/compiler/DependencyState.java
wilx: Fixed.

src/main/java/org/apache/maven/plugin/compiler/DependencyState.java
wilx: Fixed by merging the two try/catch regions.

src/main/java/org/apache/maven/plugin/compiler/DependencyState.java
wilx: I can't see any function that would be widely available.

src/main/java/org/apache/maven/plugin/compiler/DependencyState.java
wilx: It is already gated by `maven.compiler.showCompilationChanges` property.

src/main/java/org/apache/maven/plugin/compiler/DependencyState.java
wilx: It is already gated by `maven.compiler.showCompilationChanges` property.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -1605,81 +1616,6 @@ private static List<String> removeEmptyCompileSourceRoots(List<String> compileSo
...
     }
 
-    /**
-     * We just compare the timestamps of all local dependency files (inter-module dependency classpath) and the own
-     * generated classes and if we got a file which is &gt;= the build-started timestamp, then we caught a file which
-     * got changed during this build.
-     *
-     * @return {@code true} if at least one single dependency has changed.
-     */
-    protected boolean isDependencyChanged() {
```
elharo, line 1745: This method should be retained with protected access, even if it's implementation changes

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -949,7 +948,19 @@ private void executeReal() throws MojoExecutionException, CompilationFailureExce
...
+                                getOutputDirectory(),
+                                getClasspathElements(),
+                                getModulepathElements(),
+                                new DependencyState.Configuration(
+                                        fileExtensions,
+                                        getBuildStartTimeInstant().orElse(null),
+                                        staleMillis,
+                                        getLog(),
+                                        showCompilationChanges))
+                        ? "changed dependency"
```
elharo: This is too many lines for a ternary operator to be legible. Just use an if

slawekjaranowski:
@wilx please rebase

wilx:
> @wilx please rebase

Done.

slawekjaranowski:
@elharo does your comments resolved?

slawekjaranowski (approved):

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
elharo: ping

src/main/java/org/apache/maven/plugin/compiler/DependencyState.java
```diff
@@ -0,0 +1,297 @@
...
+    private static void update(MessageDigest digest, String value) {
+        digest.update(value.getBytes(StandardCharsets.UTF_8));
+        digest.update((byte) 0);
+    }
+
+    private static String toHexString(byte[] bytes) {
+        StringBuilder buffer = new StringBuilder(bytes.length * 2);
+        for (byte value : bytes) {
+            int unsigned = value & 0xFF;
+            buffer.append(Character.forDigit(unsigned >>> 4, 16));
```
elharo: I'd just use Integer.toHexString here

elharo (changes requested):
some of my original comments are answered, a couple are not

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
slawekjaranowski: I'm not sure if we need to preserve all plugin internal methods.
Maven plugin is not library so should be not used as standard dependencies.

Even more we only recommended use as plugin not dependencies
https://maven.apache.org/plugins/maven-compiler-plugin/dependency-info.html

src/main/java/org/apache/maven/plugin/compiler/DependencyState.java
slawekjaranowski: I agree here with @wilx not needed to do more complicated

slawekjaranowski:
> some of my original comments are answered, a couple are not

I hope that now you have answer for most of important questions ... 

to be honest there is not ideal code ... next time we can introduce next improvements so we should not block contributions because of nits

PR #1108, opened 2026-08-05
Keep the recompiling message at the INFO level when user explicitly requested that information

desruisseaux:
Keep the recompiling message at the INFO level when user explicitly requested that information. This pull request narrows the scope of #1089 to the case when the user did not requested that information.

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/IncrementalBuild.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/IncrementalBuild.java
@@ -318,7 +318,7 @@ final class IncrementalBuild {
     /**
      * Whether to provide more details about why a module is rebuilt.
      */
-    private final boolean showCompilationChanges;
+    final boolean showCompilationChanges;
 
     /**
      * Creates a new helper for an incremental build.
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/maven/plugin/compiler/IncrementalBuild.java
```diff
@@ -318,7 +318,7 @@ static EnumSet<Aspect> parse(final String values) {
     /**
      * Whether to provide more details about why a module is rebuilt.
      */
-    private final boolean showCompilationChanges;
+    final boolean showCompilationChanges;
```
elharo, line 321: How does the user set this? Is there a paremter somewhere?

elharo (commented):
A trace level below debug level sounds useful. I think we have defaulted to showing far too much logging that is never looked at and merely obscures real issues. 

My rule of thumb is that every log message should be interesting, actionable, and useful, especially at info level or higher. It is not enough that someone might sometime want to see the log. To log at info level we need to be confident that most people will want to see it most of the time.

More needles, less hay. :-)

src/main/java/org/apache/maven/plugin/compiler/IncrementalBuild.java
desruisseaux: Yes, this is a parameter in `CompilerMojo`. This parameter already existed in the Maven 3 plugin.

desruisseaux:
I presume that we can merge. It is an intermediate between the situation before #1089 and after.

PR #1119, opened 2026-08-28
Remove adopt-openj9

elharo:
adopt-openj9 is not available and this breaks the CI

PR #1122, opened 2026-08-29
Replace adopt-openj9 by semeru JDK distribution on GH

slawekjaranowski:

- [ ] Your pull request should address just one issue, without pulling in other changes.
- [ ] Run `mvn verify` to make sure basic checks pass.

### DIFF
--- a/.github/workflows/maven-verify.yml
+++ b/.github/workflows/maven-verify.yml
@@ -27,14 +27,13 @@ jobs:
     uses: apache/maven-gh-actions-shared/.github/workflows/maven-verify.yml@v5
     with:
       # maven-args: "-Dinvoker.parallelThreads=2" cannot do this as this generate some concurrent download issues
-      jdk-distribution-matrix: '[ "temurin", "zulu", "microsoft", "adopt-openj9" ]'
+      jdk-distribution-matrix: '[ "temurin", "zulu", "microsoft", "semeru" ]'
       jdk-matrix: '[ "8", "21", "25" ]'
       maven4-enabled: true
       verify-fail-fast: false
       matrix-exclude: '[
           { "jdk": "8", "distribution": "microsoft" },
           { "jdk": "8", "distribution": "temurin", "os": "macos-latest" },
-          { "jdk": "8", "distribution": "adopt-openj9", "os": "macos-latest" },
-          { "jdk": "25", "distribution": "adopt-openj9" },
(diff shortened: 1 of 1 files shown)


slawekjaranowski:
- #1119

.github/workflows/maven-verify.yml
```diff
@@ -27,14 +27,13 @@ jobs:
     uses: apache/maven-gh-actions-shared/.github/workflows/maven-verify.yml@v5
     with:
       # maven-args: "-Dinvoker.parallelThreads=2" cannot do this as this generate some concurrent download issues
-      jdk-distribution-matrix: '[ "temurin", "zulu", "microsoft", "adopt-openj9" ]'
+      jdk-distribution-matrix: '[ "temurin", "zulu", "microsoft", "semeru" ]'
```
elharo, line 30: OK, but why use this at all? Do we really need four here? I don't think these four are even the most common.

elharo (commented):
Running a test matrix across four different JDK distributions (**Temurin**, **Zulu**, **Microsoft**, and **Semeru**) provides limited extra coverage, and for most Java projects—including Maven plugins—it creates substantial duplicate CI work.

### 1. HotSpot vs. OpenJ9

* **HotSpot Group (`temurin`, `zulu`, `microsoft`):** All three distributions compile the exact same upstream OpenJDK C++ source code for the HotSpot Virtual Machine and Java Class Library. Their execution behavior, bytecode compilation, and core API operations are practically identical. Testing across all three yields almost zero additional coverage regarding Java language specs or plugin functionality.
* **OpenJ9 Group (`semeru`):** Uses IBM's OpenJ9 Virtual Machine engine instead of HotSpot. It has distinct garbage collection behavior, memory layout, class loading characteristics, and JVM flags. Testing on `semeru` actually tests a completely different execution engine.

To eliminate redundant CI runner time:

1. **Drop `microsoft` and `zulu**` for routine matrix runs and standardize on **`temurin`** as the primary HotSpot reference.
2. Keep **`semeru`** only if you explicitly intend to support and validate behavior under the OpenJ9 JVM engine.

.github/workflows/maven-verify.yml
slawekjaranowski: as for compiler we can detect difference eg in internal compiler arguments ....

It was added historically, now I only change deprecated  adopt-openj9 to semeru

slachiewicz (approved):

PR #1123, opened 2026-08-31
[MCOMPILER-563] Clarify that incremental compilation is not an IDE-style incremental compiler

### LINKED ISSUE #777: [MCOMPILER-563] Deprecate/rename useIncrementalCompilation
**[Jorge Solórzano](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=jorsol)** opened **[MCOMPILER-563](https://issues.apache.org/jira/browse/MCOMPILER-563?redirect=false)** and commented

The property **maven.compiler.useIncrementalCompilation** has the unfortunate name that makes it believe that Maven is using "incremental" source code compilation.

People always believe that this compiles only the single class or source code that has changed when the reality is that it only is used to detect changes and recompile the full Maven project/module.

The proposal would be to deprecate this property now and rename it in the 4.0 branch to something like **maven.compiler.rebuildDetection.**

---

**Affects:** 3.11.0

**Issue Links:**
- [MCOMPILER-515](https://issues.apache.org/jira/browse/MCOMPILER-515) Plugin is NOT incremental, so let it go

elharo:
Improves the documentation for the `incrementalCompilation` parameter and the deprecated `useIncrementalCompilation` parameter (issue #777, [MCOMPILER-563]).

The name "incremental compilation" gives the misleading impression that the plugin compiles a single changed class together with its dependents, like an IDE incremental compiler. In reality the plugin only runs a change-detection algorithm that decides whether to recompile the whole module or only the modified source files.

This is a documentation-only change (Javadoc). No Java code behavior is modified. The updated Javadoc is what gets rendered into the plugin's published parameter documentation (plugin-info `.html`).

Closes #777

### DIFF
--- a/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
@@ -680,11 +680,10 @@ public abstract class AbstractCompilerMojo implements Mojo {
 
     /**
      * Whether to enable/disable the change detection that decides when to recompile the module.
-     *
-     * <p>Despite the word "incremental" in this property name, this does <i>not</i> enable an
+     * Despite the word "incremental", this does not enable an
      * incremental compiler in the sense of an IDE. The plugin never compiles a single changed class
      * together with the classes that depend on it. It only detects changes and, depending on the
-     * configuration, recompiles the whole module or only the modified source files.</p>
+     * configuration, recompiles the whole module or only the modified source files.
      *
      * @since 3.1
      *
--- a/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
+++ b/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
@@ -586,7 +586,16 @@ public abstract class AbstractCompilerMojo implements Mojo {
     protected String outputTimestamp;
 
     /**
-     * The algorithm to use for selecting which files to compile.
+     * <b>Despite the word "incremental" in the name, this is <i>not</i> an incremental compiler
+     * in the sense of an IDE.</b> The plugin does not compile a single changed class and the classes
+     * that depend on it. It selects an algorithm used to <i>detect changes</i> and to decide whether
+     * to recompile the whole module or only some source files. In most configurations a change causes
+     * the whole module (all its source files) to be recompiled; see the {@code sources} and
+     * {@code classes} values below, which decide whether only the modified source files are recompiled
+     * or whether a change triggers a full rebuild. The plugin never performs dependency-based
+     * compilation of only the directly or transitively affected classes.
+     *
+     * <p>The algorithm to use for selecting which files to compile.
      * Values can be {@code dependencies}, {@code sources}, {@code classes}, {@code rebuild-on-change},
      * {@code rebuild-on-add}, {@code modules} or {@code none}.
      *
@@ -670,7 +679,12 @@ public abstract class AbstractCompilerMojo implements Mojo {
     protected String incrementalCompilation;
 
     /**
-     * Whether to enable/disable incremental compilation feature.
+     * Whether to enable/disable the change detection that decides when to recompile the module.
+     *
+     * <p>Despite the word "incremental" in this property name, this does <i>not</i> enable an
+     * incremental compiler in the sense of an IDE. The plugin never compiles a single changed class
+     * together with the classes that depend on it. It only detects changes and, depending on the
+     * configuration, recompiles the whole module or only the modified source files.</p>
      *
      * @since 3.1
      *


desruisseaux:
Before to change the documentation, it may be worth retesting if the issue described in #777 is still true. If I remember correctly, this parameter was intended to be incremental compilation in the IDE-style, but was not because of a bug (comparing relative paths against absolute paths).

In version 4.x of the compiler plugin, this parameter does IDE-style incremental compilation, which was apparently the original intend.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -586,7 +586,16 @@ final Charset charset() {
...
-     * The algorithm to use for selecting which files to compile.
+     * <b>Despite the word "incremental" in the name, this is <i>not</i> an incremental compiler
+     * in the sense of an IDE.</b> The plugin does not compile a single changed class and the classes
+     * that depend on it. It selects an algorithm used to <i>detect changes</i> and to decide whether
+     * to recompile the whole module or only some source files. In most configurations a change causes
+     * the whole module (all its source files) to be recompiled; see the {@code sources} and
+     * {@code classes} values below, which decide whether only the modified source files are recompiled
+     * or whether a change triggers a full rebuild. The plugin never performs dependency-based
+     * compilation of only the directly or transitively affected classes.
+     *
```
gnodet, line 597: Two inaccuracies in this block need fixing:

1. **"in most configurations a change causes the whole module to be recompiled"** — this is wrong for the default. The default (`"options,dependencies,sources"`) recompiles **only the modified source files** when a source changes. A full rebuild happens only when compiler options or a JAR dependency changed — or when annotation processors are present (Java < 23). This sentence will mislead v4 users.

2. **"The plugin never performs dependency-based compilation"** — too absolute: the `modules` algorithm delegates the per-file recompilation decision to `javac --module`, which does perform dependency-aware recompilation. The existing `<h4>Limitations</h4>` section below is already accurate; rely on that instead of a blanket "never".

```suggestion
     * <b>Despite the word "incremental" in the name, this is <i>not</i> an incremental compiler
     * in the sense of an IDE.</b> The plugin does not compile a single changed class and the classes
     * that depend on it (except when using the {@code modules} algorithm, which delegates this decision
     * to the Java compiler). It selects an algorithm used to <i>detect changes</i> and to decide whether
     * to recompile the whole module or only some source files. In the default configuration (no annotation
     * processors, Java &ge; 23), only the modified source files are recompiled; a full rebuild is triggered
     * by a compiler option change, a dependency change, or annotation processor presence.
     *
     * <p>The algorithm to use for selecting which files to compile.
```

gnodet (dismissed):
The intent of this PR is right — the parameter name is genuinely confusing and the clarification is valuable. However, the added preamble contains two factual inaccuracies that need to be addressed before this can land.

_This review was generated by an AI agent, Hermès on behalf of @gnodet._

elharo:
I would feel more comfortable with a human approval.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -586,7 +586,16 @@ final Charset charset() {
     protected String outputTimestamp;
 
     /**
-     * The algorithm to use for selecting which files to compile.
```
desruisseaux: Do not remove this line. This is the option summary. If removed, the description would start directly with a warning (_"despite the word incremental…"_) before any explanation of what the option is doing.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -586,7 +586,16 @@ final Charset charset() {
...
+     * <b>Despite the word "incremental" in the name, this is <em>not</em> an incremental compiler
+     * in the sense of an IDE.</b> The plugin does not compile a single changed class and the classes
+     * that depend on it (except when using the {@code modules} algorithm, which delegates this decision
+     * to the Java compiler). It selects an algorithm used to <i>detect changes</i> and to decide whether
+     * to recompile the whole module or only some source files. In the default configuration (no annotation
+     * processors, Java &ge; 23), only the modified source files are recompiled; a full rebuild is triggered
+     * by a compiler option change, a dependency JAR change, or annotation processor presence; see the
+     * values and the Default value section below.
```
desruisseaux: The AI-generated text is a little bit verbose for saying few (it repeats in other words the option descriptions). I also have some concerns:

* _"incremental compiler in the sense of an IDE"_ is a vague concept that depends on the IDE. Some of them just rely on `javac` behaviour, which compares timestamps like what Maven does. This is why we could claim that Maven Compiler Plugin 4.x, after the improvements done since 3.x, has an IDE-style incremental compilation, while actually it depends which IDE we compare to.
* _"except when using the `modules` algorithm, which delegates this decision to the Java compiler"_ gives the impression that `javac` checks dependencies in such case, which is not accurate. What AI claims to be a clarification actually brings confusion. The `javac` behaviour in such case is actually quite similar to the Maven behaviour. This sentence consumes space for saying something that change almost nothing for the user regarding incremental compilation.
* We do not implement yet an algorithm that check dependencies, but it has been requested by users and may be added in the future. It would be a new keyword in the list of recognized algorithms. Therefore, a wording that feel too final may be misleading.

I propose a shorter and more nuanced paragraph like below (minus my potentially broken English). Please feel free to reword:

> The incremental compilation controlled by this option is not yet as reliable as the incremental compilation provided by some IDEs, but provides an approximation based on the timestamps of source files. Many algorithms (listed below) can be combined for tuning the trade-off between overhead and reliability. The current algorithms detect only direct changes (recompiling only modified source file), but a future version of this plugin may add an option for tracking classes that depend on a modified class, as done by some IDEs.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -586,7 +586,16 @@ final Charset charset() {
...
+     *
+     * <p>The algorithm to use for selecting which files to compile.
```
desruisseaux: Remove this line here. It needs to be the first line of this javadoc.

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -670,7 +679,11 @@ final Charset charset() {
     protected String incrementalCompilation;
 
     /**
-     * Whether to enable/disable incremental compilation feature.
+     * Whether to enable/disable the change detection that decides when to recompile the module.
```
desruisseaux: Replace "module" by "sub-project".

src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java
```diff
@@ -670,7 +679,11 @@ final Charset charset() {
...
+     * Despite the word "incremental", this does not enable an
+     * incremental compiler in the sense of an IDE. The plugin never compiles a single changed class
+     * together with the classes that depend on it. It only detects changes and, depending on the
+     * configuration, recompiles the whole module or only the modified source files.
```
desruisseaux: Same comment as for the `incrementalCompilation` option: whether it is an IDE-style incremental compilation or not depends on which IDE we compare to. For example, a "build all" in NetBeans delegates to Ant, Maven or Gradle.

The word "never" is a bit strong since more reliable incremental compilation may be added in the future. I admit that it would not apply to this deprecated option, but since this option is deprecated, it may not be necessary to add this paragraph.

